PR #8, opened 2015-11-18
[component,core,datastax] clean up BackendKeyClause

udoprog:
* Rename BackendKeyCriteria -> BackendKeyClause
* Move datastax clause conversion logic into BackendKeyUtils, with some basic
  tests.
* Get rid of Datastax's QueryBuilder since its mostly in the way.
* Unify usage between data-migrate and keys shell tasks.


juruen:
:+1:

PR #22, opened 2016-01-26
Add support for FilterK aggregations

juruen:
This aggregation provides topk and bottomk filtering. Time-series are
sorted by computing the area under the curve.

It needs to be used along with a child aggregation.
e.g: "topk(3, avergage by host)"


udoprog:
Thanks, looks great!

PR #33, opened 2016-02-29
[aggregation] Use more accurate area computation for filterk

juruen:


aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/PointPairArea.java
```diff
@@ -0,0 +1,86 @@
...
+     */
+    public static double computeArea(Point a, Point b) {
+        final double x1 = a.getTimestamp();
+        final double x2 = b.getTimestamp();
+
+        if (x1 > x2) {
+            throw new IllegalArgumentException("timestamp of b needs to be greater than a");
+        }
+
+        if (x1 == x2) {
```
udoprog: One might want to perform a `long` comparison due to potentially problematic floating point comparisons.

PR #37, opened 2016-03-11
declare required Cassandra version

dimaslv:
Current version of Heroic (at least heroic-shell -P cassandra -X cassandra.seeds=<seeds> -X cassandra.configure) doesn't work with Cassandra 3, so declare correct version in dependencies.

### DIFF
--- a/source/_pages/docs/getting_started/installation.ngt
+++ b/source/_pages/docs/getting_started/installation.ngt
@@ -5,7 +5,7 @@
 </p>
 
 <ul>
-  <li><a ui-sref=".({'#': 'cassandra'})">A Cassandra cluster</a></li>
+  <li><a ui-sref=".({'#': 'cassandra'})">A Cassandra 2.x cluster</a></li>
   <li><a ui-sref=".({'#': 'elasticsearch'})">An Elasticsearch cluster</a></li>
   <li><a ui-sref=".({'#': 'kafka'})">A Kafka Cluster</a></li>
 </ul>
@@ -21,7 +21,7 @@
 </h3>
 
 <h3 id="cassandra">
-  Cassandra
+  Cassandra 2.x
 <h3>
 
 <h4>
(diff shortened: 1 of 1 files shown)


source/_pages/docs/getting_started/compile.ngt
```diff
@@ -37,5 +37,5 @@ $ mvn clean package
 </p>
 
 <pre><code language="bash">
-$ java -cp heroic-dist/target/heroic-dist-0.0.1-SNAPSHOT.jar com.spotify.heroic.HeroicService &lt;config&gt;
+$ java -cp heroic-dist/target/heroic-dist-0.0.1-SNAPSHOT-shaded.jar com.spotify.heroic.HeroicService &lt;config&gt;
```
parmus: This change is not part of the documentation update.

dimaslv: I have messed a bit with commits and branches. Now request about shaded version is in separate pull request https://github.com/spotify/heroic/pull/38.
But I have not get your comment actually.
I was following documentation and it is impossible to start Heroic with provided command without additional arguments or environment configuration. Provided fix make it possible to run  application according to documentation.

udoprog:
Thanks!

rthille:
Does anyone have a quick overview of what's broken on Cassandra 3.0.x and what it might take to get it working?  A company I work for might be interested in doing the work and contributing back...

udoprog:
@rthille according to this:
https://docs.datastax.com/en/developer/driver-matrix/doc/common/driverMatrix.html

Version 3.0 of the Java driver should be compatible with most Cassandra versions still in use. It might only be a matter of upgrading the driver version [in the repackaging](https://github.com/spotify/heroic/blob/master/repackaged/datastax/pom.xml), in the [metric backend](https://github.com/spotify/heroic/blob/master/metric/datastax/pom.xml), and check that it works.

It might be useful to adapt the existing integration tests so that they can be run against arbitrary backends (right now it's only in-memory). This should give us a suite that allow us to check for compatibility.

rthille:
@udoprog Thanks!  I'll pass this onto the guys working on Heroic (I do our Cassandra stuff).

PR #38, opened 2016-03-11
use shaded version of heroic jar

dimaslv:
only shaded version of jar could be executed without additional classpath locations


udoprog:
Thank you!

PR #40, opened 2016-03-15
change default port of nativerpc server

dimaslv:
Currently, if you would try to start new heroic cluster according to documentation, then cluster will not listen for nativerpc port and no error would be generated about reasons. The problem is that by default it bounds to port number 0. So, let's make it compliant with documentation. Otherwise, changes needed in documentation.


udoprog:
This looks reasonable. We just need to assert that the integration tests survive this.

udoprog:
Integration tests have stopped depending on this behavior. I'll merge this as soon as the tests pass. 

Thanks!

PR #43, opened 2016-04-14
[aggregation] add JsonProperty to constructor arguments

juruen:


udoprog:
Interesting, don't we need a `@JsonCreator` as well, or does Jackson 2.7.x handle this?

juruen:
Yeah, it works without @JsonCreator.

In any case, I'm going to add it for consistency.

PR #45, opened 2016-04-25
[core,component] add empty query results to response

juruen:
Our alerting system relies on heroic returning time-series with
empty values for its "data is missing" capabilities.


juruen:
@udoprog test pass but I'm not sure if this has any undesired side effects. 

Also, without this patch we have a regression in our alerting system.

PR #49, opened 2016-05-27
[all] completely remove aggregation traversal step

udoprog:
This change removed the traversal step of aggregations in favor of runtime detection.

As aggregations got more complicated with the introduction of filtering aggregations, it became much harder to pre-preemptively determine and optimize for the resulting structure of the aggregation. Filtering aggregations make this hard because it's not possible to determine 'which' result groups will be passed through until it has loaded data.

This patch removes that analysis (traversal) and replaces it with on-the-fly management of session state (see `GroupSession` use of `ConcurrentHashMap`).


juruen:
:+1:

PR #66, opened 2016-08-05
Added EditorConfig file to get indent correct.

mbrukman:
Also updated `.gitignore` to mark `.gitignore` itself as important and not
ignored, as well as the new `.editorconfig` file.


udoprog:
Thanks!

PR #67, opened 2016-08-16
[bigtable] make serialization backwards compatible

juruen:
- Upgrade tiny-serializer to 1.4.5 to configure string serialization
- Add custom string serializer to honour prevous serialization


udoprog:
Might want to change the name of `StringSerializer` to not conflict. Otherwise good stuff.

Issue #70, opened 2016-09-05
Heroic Shell status Code


jcabmora:
When the Heroic Shell ends with an error, an incorrect exit code of zero is returned.  This makes it inconvenient to use in shell scripts to automate tasks.   In the following example you can see that the exit code was `0` even though the previous command did not succeed.

```
13:33:56 heroic $ java -cp "target/*" com.spotify.heroic.HeroicShell --connect 172.16.0.13:9190
13:34:32.145 [main] INFO  com.spotify.heroic.HeroicShell - Setting up interactive shell...
13:34:32.154 [main] ERROR com.spotify.heroic.HeroicShell - Error when running shell
java.net.ConnectException: Connection refused
	at java.net.PlainSocketImpl.socketConnect(Native Method)
	at java.net.AbstractPlainSocketImpl.doConnect(AbstractPlainSocketImpl.java:350)
	at java.net.AbstractPlainSocketImpl.connectToAddress(AbstractPlainSocketImpl.java:206)
	at java.net.AbstractPlainSocketImpl.connect(AbstractPlainSocketImpl.java:188)
	at java.net.SocksSocketImpl.connect(SocksSocketImpl.java:392)
	at java.net.Socket.connect(Socket.java:589)
	at java.net.Socket.connect(Socket.java:538)
	at com.spotify.heroic.shell.RemoteCoreInterface.connect(RemoteCoreInterface.java:285)
	at com.spotify.heroic.shell.RemoteCoreInterface.commands(RemoteCoreInterface.java:254)
	at com.spotify.heroic.HeroicShell.runInteractiveShell(HeroicShell.java:197)
	at com.spotify.heroic.HeroicShell.interactive(HeroicShell.java:182)
	at com.spotify.heroic.HeroicShell.main(HeroicShell.java:116)
13:34:32.155 [main] INFO  com.spotify.heroic.HeroicShell - Closing core bridge...
13:34:32 heroic  $ echo $?
0
```

PR #71, opened 2016-09-05
HeroicShell should return exit code 1 in case of errors

jcabmora:
This addresses the problem referenced in #70 .  Please let me know if you have any questions or suggestions.  

Note: I enclosed `core.shutdown()` within a `finally` block to ensure it always runs as it currently does.   That method is currently empty but I assume that its functionality will be implemented.

### DIFF
--- a/heroic-shell/src/main/java/com/spotify/heroic/HeroicShell.java
+++ b/heroic-shell/src/main/java/com/spotify/heroic/HeroicShell.java
@@ -112,8 +112,12 @@ public class HeroicShell {
                 System.exit(1);
                 return;
             }
-
-            interactive(params, bridge);
+            try {
+                interactive(params, bridge);
+            } catch (Exception e) {
+                log.error("Error when running shell", e);
+                System.exit(1);
+            }
             System.exit(0);
             return;
         }
@@ -175,21 +179,28 @@ public class HeroicShell {
         });
     }
 
-    static void interactive(Parameters params, CoreInterface core) {
+    static void interactive(Parameters params, CoreInterface core) throws Exception {
         log.info("Setting up interactive shell...");
 
+        Exception e = null;
         try {
             runInteractiveShell(core);
-        } catch (Exception e) {
-            log.error("Error when running shell", e);
+        } catch (final Exception inner) {
+            e = inner;
         }
 
         log.info("Closing core bridge...");
 
         try {
             core.shutdown();
-        } catch (Exception e) {
-            log.error("Failed to close core bridge", e);
+        } catch (final Exception inner) {
+            if (e != null) {
+                inner.addSuppressed(e);
+            }
+            e = inner;
+        }
+        if (e != null) {
+            throw e;
         }
     }
 


udoprog:
Hey, thanks for the patch!

It currently does not take into account the result of individual running tasks, only more general errors (e.g. failed to connect to core). Is this intentional?

With your patch, a failure in the shutdown hook will currently only be logged and not cause a different exit code which can be a bit counter-intuitive. The user will see a stack-trace but the exit code will indicate that everything is fine.

As a proposal to unify your behavior you can make use of the suppression mechanism. This would cause the entirety of the error information to be available when logging the error outside of the method.

```
Exception e = null;

try {
  /* run interactive shell */
} catch(final Exception inner) {
  e = inner;
}

try {
  /* run shutdown hook */
} catch(final Exception inner) {
  if (e != null) {
    inner.addSuppressed(e);
  }

  e = inner;
}

if (e != null) {
  throw e;
}
```

I realize its a bit awkward, but its as close to a 'correct' way that I can see it barring Java providing a more convenient language mechanism for communicating suppressed exceptions.

P.S. the shutdown method for `CoreInterface` is implemented here, but only when running an embedded core: https://github.com/spotify/heroic/blob/cd0d682cf36223bc6e02607eb58a91a6e9cb2556/heroic-shell/src/main/java/com/spotify/heroic/HeroicShell.java#L173

jcabmora:
Hi!

Thanks for the feedback, and also, you are welcome!  I am happy to be able to contribute.

About the status codes, I would think that the expected behavior for a command is to return an exit code of 1 (or anything other than zero) if you could not execute a command or the command terminates unexpectedly (in this case the command being the HeroicShell, not the individual tasks that you can run with it).    I am aware that "expected"  is subjective, and it is impossible to draw a line on what is defined as such. Returning an error code for an error on an individual task executed within the HeroicShell once it is started is probably not the right thing since  it is an interactive tool and you could potentially run many tasks, and I would think that it is safe to assume that if you are going to open a HeroicShell and run multiple tasks, you probably will need to parse the output of each task you run to verify if it was successful or not.  I hope this makes sense, and if you don't agree please feel free to let me know.

I agree that if there is an Exception in the shutdown process then an exit code other than zero should be also returned.  I have applied the changes you suggested to this PR.

Please let me know if you have any questions or suggestions.

heroic-shell/src/main/java/com/spotify/heroic/HeroicShell.java
```diff
@@ -175,21 +178,31 @@ public void shutdown() throws Exception {
...
-    static void interactive(Parameters params, CoreInterface core) {
+    static void interactive(Parameters params, CoreInterface core) throws Exception {
         log.info("Setting up interactive shell...");
 
+        Exception e = null;
         try {
             runInteractiveShell(core);
-        } catch (Exception e) {
+        } catch (final Exception inner) {
             log.error("Error when running shell", e);
```
udoprog: Logs the wrong exception, also, if you log it in the catch above there is no need to log here.

heroic-shell/src/main/java/com/spotify/heroic/HeroicShell.java
```diff
@@ -175,21 +178,31 @@ public void shutdown() throws Exception {
...
+            e = inner;
         }
 
         log.info("Closing core bridge...");
 
         try {
             core.shutdown();
-        } catch (Exception e) {
+        } catch (final Exception inner) {
             log.error("Failed to close core bridge", e);
```
udoprog: Same here.

heroic-shell/src/main/java/com/spotify/heroic/HeroicShell.java
jcabmora: Holy cow, that was a silly mistake.   Sorry!   I have fixed this.

heroic-shell/src/main/java/com/spotify/heroic/HeroicShell.java
jcabmora: I have fixed this as well.  Thanks for catching this!

udoprog:
Hey, final thing; you only need to log the exception once (at the `System.exit` point) when using suppressed exceptions since the one thrown exception contains the suppressed if it was thrown earlier. Otherwise the suppression mechanism doesn't actually do anything since the exception is just discarded.

jcabmora:
Hi, Sorry for the super delayed response.  Thanks for reviewing my changes. I have updated the log statements per your comments.   If there is anything that you think needs to be updated, let me know.

PR #73, opened 2016-09-12
fix links to idea/eclipse directories

mattnworb:
When viewing https://github.com/spotify/heroic, Github renders the links
to `idea` and `eclipse` as links to `https://github.com/spotify/idea`
etc. Fix this by linking to the README instead.


mattnworb:
@juruen another one for you 😄

udoprog:
I wonder how @codecov-io reasons that this will decrease coverage.

mattnworb:
I think my original commit was a few commits behind master

PR #78, opened 2016-09-18
[parser] relax SimpleString and fix precedence (fixes #76)

### LINKED ISSUE #76: Issue with Expression parsing
Hi guys,

I am using the `metadata-fetch` task from HeroicShell.  I ran into some inconsistencies in how the filters are parsed that I believe are caused by how the HeroicQuery grammar is defined.  
As an example, supposed these two keys exist: `4cf41d4b-a5d2-420e-b008-644f0b3d7832` and `c89bf1ee-32ad-44c9-9641-5ee7fe3cac17`. If I run this command I can successfully get the series related to key `c89bf1ee-32ad-44c9-9641-5ee7fe3cac17`:
`metadata-fetch $key = c89bf1ee-32ad-44c9-9641-5ee7fe3cac17`
However, if I run the same for the other key, I get a `com.spotify.heroic.grammar.ParseException`:
`metadata-fetch $key = 4cf41d4b-a5d2-420e-b008-644f0b3d7832`
I found that you have to escape the whole expression to be able to make it work:
`metadata-fetch '$key = "4cf41d4b-a5d2-420e-b008-644f0b3d7832"'`

This problem only happens with expressions that begin with a numeric.  I believe the problem is caused  when [this Rule](https://github.com/spotify/heroic/blob/master/heroic-parser/src/main/antlr4/com/spotify/heroic/grammar/HeroicQuery.g4#L55) is applied, `expr` is matched first for an `ExpressionInteger` and then instead of finding `EOF` if finds the rest of the string.  When quotes are applied it correctly matches `QuotedString`.  I know this can be worked around, but probably having consistency in the syntax would be good.

Regards,

Jorge

udoprog:
This changes the precedence of the `SimpleString` lexer rule to happen after
`Integer`, `Float`, and `Duration`.

It also allows `SimpleString` to start with a numeric component, but because of
the change in precedence the above will match before if applicable.

This allows the following expression to work (as referenced in #76):
```
$key = 4cf41d4b-a5d2-420e-b008-644f0b3d7832
```

### DIFF
--- a/heroic-parser/src/main/antlr4/com/spotify/heroic/grammar/HeroicQuery.g4
+++ b/heroic-parser/src/main/antlr4/com/spotify/heroic/grammar/HeroicQuery.g4
@@ -188,11 +188,6 @@ QuotedString : '"' StringCharacters? '"' ;
 
 Identifier : [a-zA-Z] [a-zA-Z0-9]* ;
 
-// strings that do not have to be quoted
-SimpleString
-    : [a-zA-Z] [a-zA-Z0-9:/_\-\.]*
-    ;
-
 Duration
     : Minus? Integer Unit
     ;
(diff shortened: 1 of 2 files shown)


juruen:
LGTM!

PR #80, opened 2016-09-22
[aggregation/simple] fix api regression in filtering aggregations (fixes #79)

### LINKED ISSUE #79: Filtering aggregations should not take child aggregations any longer
29d83a7b1247e8a91361cc2533ae3f8a971458f2 removed `of` as a parameter to filtering aggregations. This is an API regression and should be replaced with a check.

udoprog:
@juruen Here is the fix, was not able to get the request mocking up and running yet but it is in progress for API tests.


juruen:
:+1:

PR #83, opened 2016-10-04
Log schema errors

juruen:


udoprog:
You might want to do this in the consumer thread instead?

PR #84, opened 2016-10-04
[all] implement watchdogs and set error threshold for metric/bigtable

juruen:
Conflicts:
	metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableBackend.java


udoprog:
Watchdogs are currently very opinionated and might mess things up for users without them having the ability to work around it.

Before releasing this, I'd like to be to disable watchdogs in the heroic configuration (or that they are disabled by default). There should also be a little bit of documentation about this feature in the README that mentions how to disable it if it does mess something up for you.

udoprog:
I've re-based the branch. It still applies cleanly to master which have had a few updates in how things are tested.

PR #85, opened 2016-10-06
[component,core] enable shift_range by default and allow disabling of features

udoprog:
Enable `shift_range` by default.

Add a `disabled` field to query and configuration which will cause the specified feature to be disabled for that query.

Clarify that features should be named to be enabled (not disabled) to avoid future complexity.

This could also be implemented so that the set of features is specified using operators as prefix, like:

Features are applied in the following order:

`(((default_features + configuration_features) - configuration_disabled) + query_features) - query_disabled`

`+` being joining two sets and `-` means subtracting the right hand side set from the left hand side.

#### Query

```javascript
{
  /* query */,
  "features": [
    "com.spotify.heroic.bar"
  ],
  "disabled": [
    "com.spotify.heroic.baz"
  ]
}
```

#### Configuration

```yaml
features:
  - com.spotify.heroic.bar

disabled:
  - com.spotify.heroic.baz
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -17,7 +17,7 @@ install:
   - tools/install-repackaged
 
 script:
-  - mvn -nsu -T 1.5C -D environment=test -P '!findbugs' verify
+  - mvn -nsu -T 2 -D environment=test -P '!findbugs,codecov' verify
 
 after_success:
   - bash <(curl -s https://codecov.io/bash)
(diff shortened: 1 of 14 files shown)


udoprog:
This could also be implemented so that the set of features is specified using operators as prefix, like:

```json
{
  /* query */,
  "features": [
    "com.spotify.heroic.bar",
    "-com.spotify.heroic.baz"
  ]
}
```

Configuration:

```yaml
features:
    "-com.spotify.heroic.baz"
```

juruen:
@udoprog Could you please fix the conflict so we can merge this ASAP? Thank you!

PR #91, opened 2016-10-09
[parser] use visitor when parsing

udoprog:
This uses the visitor pattern to parse the grammar instead of maintaining an internal stack.

See http://jakubdziworski.github.io/java/2016/04/01/antlr_visitor_vs_listener.html

Instead of relying on things being parsed in order and validating the stack as it is being popped, this makes use of references generated by the grammar.

Expressions like the following:
```java
FunctionExpression function = pop(ctx, FunctionExpression.class);
// vs
FunctionExpression function = ctx.function().accept(functionVisitor);
```

Or when maintaining lists of data, we no longer need to use tokens to iterate over results:

```java
final List<Expression> arguments = popUntil(ctx, END_OF_LIST_MARK);
// vs
final List<Expression> arguments = ctx.expr().stream().map(e -> e.accept(expressionVisitor)).collect(Collectors.toList());
```


hexedpackets:
Closing due to age, feel free to re-open if this is still an issue.

PR #93, opened 2016-10-09
[dist] skip shading when environment=test

udoprog:
This will disable shading for `dist` when `environment=test` is specified which is specifically designed for running all available test suites, and shading should be an unnecessary step.

This will not affect regular packaging builds (`mvn clean package`).

This build should give an indication how how faster testing will be without this step.


udoprog:
`Elapsed time 11 min 6 sec` to `Elapsed time 9 min 59 sec`.

PR #94, opened 2016-10-09
[travis] build maven modules in parallel

udoprog:


udoprog:
Down to `Elapsed time 7 min 57 sec`.

PR #101, opened 2016-10-10
[core] actually start the internal life cycles and test it

udoprog:
This fixes a startup regression introduced in 5094210.

The issue was caused by the internal started future never being resolved, which prevents the initial cluster refresh from being called as is done here:
https://github.com/spotify/heroic/blob/master/heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java#L306


parmus:
Looks good to me. I can confirm that I can now start and query Heroic again without crashes.

PR #107, opened 2016-10-11
[all] fix limiting of results (again)

udoprog:
Also add a test to make sure it really does not happen again.


juruen:
@udoprog could you please fix the conflict? Thanks!

PR #112, opened 2016-10-17
Fix ScrollTransform<T>

parmus:
ScrollTransform<T> have a number of minor bugs:

- Duplicates count towards to limit making the actual result unpredictable
- Limit check is off-by-one
- If the number of results is exactly one more than the limit, isLimited is incorrectly set to false, because the limit check happens in the beginning of the loop

This patch addresses all of the above.

### DIFF
--- a/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/AbstractElasticsearchMetadataBackend.java
+++ b/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/AbstractElasticsearchMetadataBackend.java
@@ -109,15 +109,18 @@ public abstract class AbstractElasticsearchMetadataBackend extends AbstractElast
             final SearchHit[] hits = response.getHits().getHits();
 
             for (final SearchHit hit : hits) {
+                final T convertedHit = converter.apply(hit);
+
+                if (!results.add(convertedHit)) {
+                    duplicates += 1;
+                } else {
+                    size += 1;
+                }
+
                 if (limit.isGreater(size)) {
+                    results.remove(convertedHit);
                     return async.resolved(new LimitedSet<>(limit.limitSet(results), true));
                 }
-
-                if (!results.add(converter.apply(hit))) {
-                    duplicates += 1;
-                }
-
-                size += 1;
             }
 
             if (hits.length == 0) {
--- a/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/AbstractElasticsearchMetadataBackend.java
+++ b/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/AbstractElasticsearchMetadataBackend.java
@@ -86,12 +86,15 @@ public abstract class AbstractElasticsearchMetadataBackend extends AbstractElast
             final Supplier<AsyncFuture<SearchResponse>> scroller =
                 () -> bind(c.prepareSearchScroll(scrollId).setScroll(SCROLL_TIME).execute());
 
-            return scroller.get().lazyTransform(new ScrollTransform<>(limit, scroller, converter));
+            return scroller.get().lazyTransform(
+                new ScrollTransform<>(async, limit, scroller, converter)
+            );
         });
     }
 
     @RequiredArgsConstructor
-    class ScrollTransform<T> implements LazyTransform<SearchResponse, LimitedSet<T>> {
+    public static class ScrollTransform<T> implements LazyTransform<SearchResponse, LimitedSet<T>> {
+        private final AsyncFramework async;
         private final OptionalLimit limit;
         private final Supplier<AsyncFuture<SearchResponse>> scroller;
 
--- a/heroic-elasticsearch-utils/src/test/java/com/spotify/heroic/elasticsearch/ScrollTransformTest.java
+++ b/heroic-elasticsearch-utils/src/test/java/com/spotify/heroic/elasticsearch/ScrollTransformTest.java
@@ -1,0 +1,172 @@
+package com.spotify.heroic.elasticsearch;
+
+import com.spotify.heroic.common.OptionalLimit;
+import com.spotify.heroic.elasticsearch.AbstractElasticsearchMetadataBackend.LimitedSet;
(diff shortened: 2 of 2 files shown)


parmus:
@udoprog , @juruen : Please review

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/AbstractElasticsearchMetadataBackend.java
```diff
@@ -106,15 +107,16 @@ public AbstractElasticsearchMetadataBackend(final AsyncFramework async, final St
...
-
                 if (!results.add(converter.apply(hit))) {
                     duplicates += 1;
+                } else {
+                    size += 1;
                 }
 
-                size += 1;
+                if (limit.isGreater(size)) {
+                    results.remove(converter.apply(hit));
```
udoprog: Only need to convert once.

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/AbstractElasticsearchMetadataBackend.java
```diff
@@ -86,12 +86,13 @@ public AbstractElasticsearchMetadataBackend(final AsyncFramework async, final St
...
                 () -> bind(c.prepareSearchScroll(scrollId).setScroll(SCROLL_TIME).execute());
 
-            return scroller.get().lazyTransform(new ScrollTransform<>(limit, scroller, converter));
+            return scroller.get().lazyTransform(new ScrollTransform<>(async, limit, scroller, converter));
         });
     }
 
     @RequiredArgsConstructor
-    class ScrollTransform<T> implements LazyTransform<SearchResponse, LimitedSet<T>> {
+    static public class ScrollTransform<T> implements LazyTransform<SearchResponse, LimitedSet<T>> {
```
udoprog: JLS order: http://checkstyle.sourceforge.net/config_modifier.html

heroic-elasticsearch-utils/src/test/java/com/spotify/heroic/elasticsearch/index/ScrollTransformTest.java
```diff
@@ -0,0 +1,340 @@
...
+        final SearchResponse searchResponse = mock(SearchResponse.class);
+        final SearchHits searchHits = mock(SearchHits.class);
+        when(searchResponse.getHits()).thenReturn(searchHits);
+        OngoingStubbing<SearchHit[]> stub = when(searchHits.getHits());
+        for(final SearchHit[] page: pages){
+            stub = stub.thenReturn(page);
+        }
+
+        Supplier<AsyncFuture<SearchResponse>> scroller = (Supplier<AsyncFuture<SearchResponse>>) mock(Supplier.class);
+        when(scroller.get()).thenReturn(new AsyncFuture<SearchResponse>() {
```
udoprog: Mock `AsyncFuture<SearchResponse>` with:
```
@Mock
AsyncFuture<SearchResponse> response;
```

And in `@Before`:
```
doAnswer(...).when(response).transform(...)
```
See: https://testing.googleblog.com/2014/03/whenhow-to-use-mockito-answer.html

heroic-elasticsearch-utils/src/test/java/com/spotify/heroic/elasticsearch/index/ScrollTransformTest.java
```diff
@@ -0,0 +1,340 @@
...
+            @Override
+            public SearchResponse get(long l, TimeUnit timeUnit) throws InterruptedException, ExecutionException, TimeoutException {
+                return null;
+            }
+        });
+
+        return scroller;
+    }
+
+    public AbstractElasticsearchMetadataBackend.ScrollTransform<Integer> createScrollTranform(Integer limit,
```
udoprog: Spelling: Tranform

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/AbstractElasticsearchMetadataBackend.java
udoprog: Not really arguing for performance, but readability yes. It makes it clearer that they are the same instance (or equivalent, in this case).

heroic-elasticsearch-utils/src/test/java/com/spotify/heroic/elasticsearch/ScrollTransformTest.java
```diff
@@ -0,0 +1,167 @@
...
+        final LimitedSet<Integer> limitedSet =
+            new LimitedSet<>(new HashSet<>(), limit != null);
+        Stream<SearchHit> stream = Arrays.stream(pages)
+              .map(Arrays::stream)
+              .reduce(Stream.empty(), Stream::concat);
+        if (limit!=null){
+            stream = stream.limit(limit);
+        }
+        stream.map(Object::hashCode)
+              .forEach(limitedSet.getSet()::add);
```
udoprog: Containers are generally designed to be immutable. I don't consider it a safe assumption that the set returned by this getter is necessarily the same (it happens to be now) or that it is mutable. It also looks like it should be possible to build the `HashSet` first, then constructing a `LimitedSet` out of that.

heroic-elasticsearch-utils/src/test/java/com/spotify/heroic/elasticsearch/ScrollTransformTest.java
```diff
@@ -0,0 +1,167 @@
...
+            stub = stub.thenReturn(page);
+        }
+    }
+
+    public ScrollTransform<Integer> createScrollTransform(
+        Integer limit,
+        Supplier<AsyncFuture<SearchResponse>> scroller
+    ){
+        final OptionalLimit optionalLimit = OptionalLimit.of(limit);
+        return new ScrollTransform<>(async, optionalLimit, scroller, Object::hashCode);
```
udoprog: [Object#hashCode()](https://docs.oracle.com/javase/8/docs/api/java/lang/Object.html#hashCode--) is not guaranteed to be unique in the local runtime environment. It would be safer to use `Function.identity()` and do equals comparisons on the actual SearchHits. These should be unique since they are individual mocks.

PR #115, opened 2016-10-22
Improved Tracing Framework

udoprog:
This change introduces the `Tracing` type and extends the capabilities of `QueryTrace`.

The new `Tracing` type encapsulated if tracing is enabled or not. And when tracing is disabled, performs no-ops in place of the actual tracing operations.

Consider the following example:

``` java
public QueryTrace tracedMethod() {
  final Tracing tracing = Tracing.fromBoolean(true);
  final QueryTrace.NamedWatch parentWatch = tracing.watch(QueryTrace.identifier("parent"));

  final QueryTrace.Joiner parentJoiner = parentWatch.joiner();

  {
    final QueryTrace.NamedWatch childWatch = parentWatch.watch(QueryTrace.identifier("one"));
    /* do something that takes time */
    parentJoiner.addChild(childWatch.end());
  }

  {
    final QueryTrace.NamedWatch childWatch = parentWatch.watch(QueryTrace.identifier("two"));
    /* do something that takes time */
    parentJoiner.addChild(childWatch.end());
  }

  return parentJoiner.result();
}
```

When using `Tracing.fromBoolean(false)` in the above code, all the code involving tracing would become no-ops (apart from the dynamic dispatch). This makes it easier to write code where potentially expensive tracing can be disabled.

`Tracing` is serialized as JSON `true` or `false` (depending on if it is enabled), making it a drop-in replacement for the existing tracing option, which is a `Boolean`.

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/metric/QueryTrace.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/metric/QueryTrace.java
@@ -22,128 +22,517 @@
 package com.spotify.heroic.metric;
 
 import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.annotation.JsonProperty;
+import com.fasterxml.jackson.annotation.JsonTypeName;
+import com.fasterxml.jackson.core.JsonGenerator;
+import com.fasterxml.jackson.core.JsonParser;
+import com.fasterxml.jackson.databind.DeserializationContext;
+import com.fasterxml.jackson.databind.JsonDeserializer;
+import com.fasterxml.jackson.databind.JsonMappingException;
+import com.fasterxml.jackson.databind.JsonSerializer;
+import com.fasterxml.jackson.databind.SerializerProvider;
+import com.fasterxml.jackson.databind.annotation.JsonDeserialize;
+import com.fasterxml.jackson.databind.annotation.JsonSerialize;
 import com.google.common.base.Stopwatch;
 import com.google.common.collect.ImmutableList;
 import lombok.Data;
 
+import java.io.IOException;
 import java.io.PrintWriter;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
-@Data
-public class QueryTrace {
-    private final Identifier what;
-    private final long elapsed;
-    private final List<QueryTrace> children;
+@JsonSerialize(using = QueryTrace.Serializer.class)
+@JsonDeserialize(using = QueryTrace.Deserializer.class)
+public interface QueryTrace {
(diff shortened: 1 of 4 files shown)


parmus:
I'm not sure I fully understand the trace functionality, but this change looks good to me.

+1

udoprog:
@parmus you can see the [tracing-type](https://github.com/spotify/heroic/tree/tracing-type) branch for how this is incorporated into the codebase.

heroic-component/src/main/java/com/spotify/heroic/metric/QueryTrace.java
```diff
@@ -22,128 +22,517 @@
...
+    /**
+     * @deprecated use {@link #what()}
+     */
+    @Deprecated
+    Identifier getWhat();
+
+    /**
+     * @deprecated use {@link #children()}
+     */
+    @Deprecated
```
gabrielgerhardsson: What is the reasoning behind deprecating these methods and replacing them? Shorter method name?

udoprog: Removing those methods so they won't get automatically picked up by ObjectMapper and identified asfields when serialized to JSON (note the `@JsonIgnore`) annotations in this file.

There are additional commits in another branch (for future PRs) that removes them, but this is the first step.

heroic-component/src/test/java/com/spotify/heroic/metric/QueryTraceTest.java
```diff
@@ -0,0 +1,113 @@
...
+    @Test
+    public void namedWatch() {
+        final QueryTrace.Identifier identifier = QueryTrace.identifier("foo");
+        final QueryTrace.NamedWatch watch =
+            new QueryTrace.ActiveNamedWatch(identifier, Stopwatch.createStarted());
+        final QueryTrace trace = watch.end();
+
+        assertTrue(trace instanceof QueryTrace.ActiveTrace);
+        final QueryTrace.ActiveTrace activeTrace = (QueryTrace.ActiveTrace) trace;
+        assertEquals(identifier, activeTrace.getWhat());
```
udoprog: The Matchers are in `heroic-dist` and not accessible to this module. There is a future refactoring planned to modularize better so that hey can be shared.

heroic-component/src/main/java/com/spotify/heroic/metric/QueryTrace.java
gabrielgerhardsson: Ah I see!

heroic-component/src/test/java/com/spotify/heroic/metric/QueryTraceTest.java
gabrielgerhardsson: Ok!

Issue #116, opened 2016-10-24
Diff aggregations (deltas)


mykolasmith:
Hi,

Are there any plans for a rate of change aggregator (deltas)? KairosDB has a "diff" aggregator that performs no sampling. We usually use it as the first aggregator in a chain to feed into downsampling aggregators (min, max, avg, etc.)

It seems like Heroic uses a concurrent bucketing pattern for `SamplingAggregation`, so it seems like a `FilterAggregation` would be more appropriate, since you would need to keep track of the state of the last point. Something like the `FilterKAreaStrategy` might work, which implements  `MetricCollection`, perhaps containing the two points at a time... e.g.:

```
private double computeDiff(MetricCollection metricCollection) {
    final List<Point> metrics = metricCollection.getDataAs(Point.class);
    return metrics.get(1).getValue() - metrics.get(0).getValue()
}
```

Am I on the right track?

udoprog:
Hey,

You are on the right track conceptually, but `FilteringAggregation` is not a great base for what you are trying to accomplish. I'd suggest investigating implementing a completely custom aggregation which wraps around `Empty`/`EmptyInstance`, and performs the delta's in the `result` method.

I realize this is what filtering aggregations do, but they are a bit more opinionated right now. What you do could be generalized into another aggregation archetype, maybe one that filtering should inherit from.

Good luck, and don't hesitate to follow up if you have more questions.

mykolasmith:
@udoprog Thanks, I'll keep you posted w/ progress & questions.

mykolasmith:
@udoprog I've created `Delta / DeltaInstance` classes that extend `Empty / EmptyInstance`. It seems that `CollectorSession` in `EmptyInstance` is a private static final class. I'm not sure how to override the `result` function of a private class from within an extending class. Apologies, I am not a Java developer =( Would you be so kind as to set up the next few lines?

```public class DeltaInstance extends EmptyInstance {
    @JsonCreator
    public DeltaInstance() {
        super();
    }
}```

**Comment by udoprog** [comment]
There´s no need to extend EmptyInstance, you can keep it as a field of `DeltaInstance`.

```java
public class DeltaInstance {
    private final EmptyInstance inner = Empty.INSTANCE;

    /* implement methods and delegate calls to inner */
}
```

This is basically how it's used in `FilteringAggregation` here:
[...]

PR #120, opened 2016-10-26
Delta aggregation

mykolasmith:
Adds delta aggregation (i.e. "diff" in KairosDB or "rate" in OpenTSDB; open to alternative names). There are no arguments, leaving it up to the user to decide whether to chain it as an input to a `SamplingAggregation`, or to diff the output for any other `Aggregation`. Example:

```
{
  "range": {"type": "relative", "unit": "HOURS", "value": 2},
  "filter": ["and", ["key", "foo"], ["=", "foo", "bar"], ["+", "role"]],
  "aggregation": {"type": "delta"},
  "groupBy": ["site"]
}
```

The above would return the difference of each point from the last within the sample. ~~The first point in a difference is always 0.~~ (EDIT: The First point sampled is truncated and the first _returned_ point is the difference from the last point).

This is my first time writing Java, so please go easy! I understand it might not be the most idiomatic code, but I did my best to use patterns found in the other aggregators (stream, map, etc.). Thanks for looking.

### DIFF
--- a/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/Delta.java
+++ b/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/Delta.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright (c) 2015 Spotify AB.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.heroic.aggregation.simple;
+
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.spotify.heroic.aggregation.Aggregation;
+import com.spotify.heroic.aggregation.AggregationContext;
+import com.spotify.heroic.aggregation.AggregationInstance;
+import lombok.Data;
+
+@Data
+public class Delta implements Aggregation {
+    public static final String NAME = "delta";
+
+    @JsonCreator
+    public Delta() { }
+
+    @Override
+    public AggregationInstance apply(
+        final AggregationContext aggregationContext
+    ) {
+        return new DeltaInstance();
+    }
+}
--- a/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/DeltaInstance.java
+++ b/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/DeltaInstance.java
@@ -1,0 +1,149 @@
+/*
+ * Copyright (c) 2015 Spotify AB.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.heroic.aggregation.simple;
+import com.spotify.heroic.aggregation.AggregationInstance;
+import com.spotify.heroic.aggregation.AggregationResult;
+import com.spotify.heroic.aggregation.AggregationSession;
+import com.spotify.heroic.aggregation.AggregationOutput;
+import com.spotify.heroic.aggregation.EmptyInstance;
+import com.spotify.heroic.common.DateRange;
+import com.spotify.heroic.common.Series;
+import com.spotify.heroic.metric.MetricCollection;
+import com.spotify.heroic.metric.MetricGroup;
+import com.spotify.heroic.metric.Event;
+import com.spotify.heroic.metric.Payload;
+import com.spotify.heroic.metric.MetricType;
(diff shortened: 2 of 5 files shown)


aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/Delta.java
```diff
@@ -0,0 +1,45 @@
...
+import com.spotify.heroic.aggregation.AggregationInstance;
+import lombok.Data;
+
+@Data
+public class Delta implements Aggregation {
+    public static final String NAME = "delta";
+
+    @JsonCreator
+    public Delta() {
+        super();
```
udoprog: No need to call `super()` since it currently doesn't extend anything.

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/DeltaInstance.java
```diff
@@ -0,0 +1,142 @@
...
+            final Map<String, String> key, final Set<Series> series, final List<Event> values
+        ) {
+            this.childSession.updateEvents(key, series, values);
+        }
+
+        @Override
+        public void updatePayload(
+            final Map<String, String> key, final Set<Series> series, final List<Payload> values
+        ) {
+            this.childSession.updatePayload(key, series, values);
```
udoprog: Leave this and others empty if you don't intend to support any other metric types. This will cause it to ignore data which is of the wrong type.

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/DeltaInstance.java
```diff
@@ -0,0 +1,142 @@
...
+    @Override
+    public boolean distributable() {
+        return false;
+    }
+
+    public List<Point> computeDiff(List<Point> points) {
+        final List<Point> result = new ArrayList<Point>();
+        double lastValue = points.get(0).getValue();
+        for (int i = 0; i < points.size(); i++) {
+            Point thisPoint = points.get(i);
```
udoprog: In general there's no guarantee that a list implementation has an efficient `get(int)` method, see [LinkedList](https://docs.oracle.com/javase/8/docs/api/java/util/LinkedList.html) for an example. Use an iterator (preferred) or a foreach.

If you want inspiration I usually do the following when I have a previous value that I want to use in the operation:
```java
final Iterator<Point> it = points.iterator();

if (!it.hasNext()) {
    return Collections.emptyList();
}

Point previous = it.next();

while (it.hasNext()) {
    final Point current = it.next();
    /* do your thing */
    previous = current;
}
```

aggregation/simple/src/test/java/com/spotify/heroic/aggregation/simple/DeltaTest.java
```diff
@@ -0,0 +1,49 @@
...
+import com.spotify.heroic.aggregation.AggregationSession;
+import com.spotify.heroic.common.DateRange;
+import com.spotify.heroic.metric.Point;
+import org.junit.Test;
+
+import java.util.ArrayList;
+
+import static org.junit.Assert.assertEquals;
+
+public class DeltaTest {
```
udoprog: Good stuff, but consider adding a test to [AbstractClusterQueryIT](https://github.com/spotify/heroic/blob/master/heroic-dist/src/test/java/com/spotify/heroic/AbstractClusterQueryIT.java), that will make sure your aggregation works as expected over a cluster (i.e. it is not distributable).

udoprog:
Relates to #116

udoprog:
Regarding the build failure, you can run your local build like so to test it before pushing:
```
mvn -D environment=test clean verify
```

mykolasmith:
@udoprog I made the changes you requested, sans the AbstractClusterQueryIT test. I was unable to figure out how to make a query using only the delta aggregator with no arguments. Does the HQL parser assume that every aggregation has some sort of unit / value?

udoprog:
What issue did you have?

mykolasmith:
@udoprog Ok, so if I understand this correctly, the points being written are:

```
    [10, 1D],
    [20, 4D],
    [30, 2D]
```

Therefore, the delta aggregation should return `[[20, 3D], [30, -2D]]`.  I am running the following block of code:

```
    @Test
    public void deltaQueryTest() throws Exception {
        final QueryResult result =
            query("delta()", builder -> {
                builder.features(Optional.empty());
            });

        final Set<MetricCollection> m = getResults(result);
        final List<Long> cadences = getCadences(result);

        assertEquals(ImmutableList.of(-1L, -1L), cadences);
        assertEquals(ImmutableSet.of(points().p(20, 3D).p(30, -2D).build(),
            points().p(20, 3D).p(30, -2D).build()), m);
    }
```

And receiving the following...

```deltaQueryTest(com.spotify.heroic.JvmClusterQueryIT)  Time elapsed: 0.064 sec  <<< ERROR!
com.spotify.heroic.grammar.ParseException: 1:6: unexpected token: )```

What is the correct way to call an aggregation with no arguments? From the docs: `"Aggregations are expressed as function calls, or typed documents."`

**Comment by mykolasmith** [comment]
@udoprog I assume you meant without parentheses. The result is that it appears that no aggregation was run:

Using both: `"delta"` and `"delta by *"`

```  GrpcClusterQueryIT>AbstractClusterQueryIT.deltaQueryTest:174
expected:<[MetricCollection(type=points, data=[Point(timestamp=20, value=3.0), Point(timestamp=30, value=-2.0)])]>

but was:<[MetricCollection(type=points, data=[Point(timestamp=10, value=1.0), Point(timestamp=30, value=2.0)]), MetricCollection(type=points, data=[Point(timestamp=10, value=1.0), Point(timestamp=20, value=4.0)])]>
```

udoprog:
I opened OdenTech/heroic#1 and tried to describe why I think it is correct. Please have a look and see if it makes sense to you.

Also, please feel free to squash the changes into this PR (including my contribution). It makes for an easier to read history.

mykolasmith:
@udoprog The expected results make sense to me, but the tests are still failing to produce the correct values. Does this mean something is wrong with my aggregation?

```deltaQueryTest:174 expected:<[MetricCollection(type=points, data=[Point(timestamp=30, value=1.0)]), MetricCollection(type=points, data=[Point(timestamp=20, value=3.0)])]> but was:<[MetricCollection(type=points, data=[Point(timestamp=10, value=1.0), Point(timestamp=30, value=2.0)]), MetricCollection(type=points, data=[Point(timestamp=10, value=1.0), Point(timestamp=20, value=4.0)])]>```

```distributedDeltaQueryTest:185 expected:<[MetricCollection(type=points, data=[Point(timestamp=20, value=3.0), Point(timestamp=30, value=-2.0)])]> but was:<[MetricCollection(type=points, data=[Point(timestamp=10, value=1.0), Point(timestamp=20, value=4.0), Point(timestamp=30, value=2.0)])]>```

udoprog:
LGTM

Final thing: please clean up the commits. Rebase and squash out some of the intermediate mistakes.You should probably only end up with one commit. Maybe two if you'd like to separate the adding of the test cases. It makes it easier to track what caused a particular change in the future for us and manage history.

udoprog:
Also pinging @juruen and @parmus to see if they are OK with merging this now.

mykolasmith:
Squashed everything into a single commit and merged in master.

PR #123, opened 2016-10-31
Report when a time series is filtered out during ingest

gabrielgspotify:
Report dropped-by-filter when a filter drops a time series during ingest. This fixes a "XXX:" comment.
Also updated tests.

Is this the right way of doing this? Any comments?


udoprog:
Hey!

That is correct.

Fix the formatting (I recommend you use the code style provided here for future reference: https://github.com/spotify/heroic/tree/master/idea) and the build should pass.

PR #124, opened 2016-10-31
Report when a time series is filtered out during ingest, try#2

gabrielgerhardsson:
Report dropped-by-filter when a filter drops a time series during ingest. This fixes a "XXX:" comment.
Also updated tests.


udoprog:
Thanks! Decrease in test coverage is due to the added lines in the semantic module. Ignoring them for now.

PR #128, opened 2016-11-09
[statistics/semantic] add pending counter

udoprog:
This adds pending counters to instrumented futures.

The pending counter is incremented when the future is created, and decremented when it finishes. This gives an indication of how many pending operations there are for a given type of future.


juruen:
:+1:

PR #130, opened 2016-11-14
Bumped Bigtable client library to 0.9.4

parmus:
Bigtable client library 0.9.4 downgraded protobuf as well, which is why getColumnFamiliesMap() change to getColumnFamilies()


udoprog:
:+1:

juruen:
:+1:

Issue #126, opened 2016-11-02
Implement slowlog


udoprog:
A `slowlog` is a log that prints information about queries which are slow.

It is typically used to figure out what is adding stress to the system, and is a feature that is required to find them.

@juruen has done some experimentation regarding this, so I'm assigning this issue to him.

dmichel1:
There is a general query log that can be used for this purpose.

PR #131, opened 2016-11-14
Query log - configurable logging of big queries

gabrielgerhardsson:
This adds the possibility to log queries, including information such as ip:port, elapsed time, number of resulting data points, full request.
Define a new Logger called "query.log" to get the output to a separate file. Log level for the query logging is "debug".
Query logging is disabled by default. Use logQueries in the metrics section to enable it.
Logging can be filtered on the amount of data points being returned by the query. Use logQueriesThresholdDataPoints in the metrics section to define lower threshold. Default threshold is 1000000. Sample configuration:

```
metrics:
  logQueries: true
  logQueriesThresholdDataPoints: 1000000
```

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/QueryOriginContext.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/QueryOriginContext.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright (c) 2016 Spotify AB.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.heroic;
+
+import java.util.UUID;
+import lombok.AllArgsConstructor;
+import lombok.Data;
+
+@AllArgsConstructor
+@Data
+public class QueryOriginContext {
+    private final String remoteAddr;
+    private final String remoteHost;
+    private final int    remotePort;
+    private final String remoteUserAgent;
+    private final String remoteClientId;
+    private final UUID   queryId;
+    private final String queryString;
+
+    public static QueryOriginContext empty() {
+        return new QueryOriginContext("", "", 0, "", "", UUID.randomUUID(), "");
+    }
+
+    public static QueryOriginContext of(final String  remoteAddr,
+                                        final String  remoteHost,
+                                        final int     remotePort,
+                                        final String  remoteUserAgent,
+                                        final String  remoteClientId,
+                                        final String  queryString) {
+        return new QueryOriginContext(remoteAddr, remoteHost, remotePort, remoteUserAgent,
+                                      remoteClientId, UUID.randomUUID(), queryString);
+    }
+}
--- a/heroic-component/src/main/java/com/spotify/heroic/metric/QueryTrace.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/metric/QueryTrace.java
@@ -25,17 +25,30 @@ import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.base.Stopwatch;
 import com.google.common.collect.ImmutableList;
+import lombok.AllArgsConstructor;
 import lombok.Data;
 
 import java.io.PrintWriter;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
+@AllArgsConstructor
 @Data
 public class QueryTrace {
     private final Identifier what;
     private final long elapsed;
     private final List<QueryTrace> children;
+
+    private long preAggregationSampleSize;
+    private long numSeries;
+
+    public QueryTrace(final Identifier what, final long elapsed, final List<QueryTrace> children) {
+        this(what, elapsed, children, 0, 0);
+        for (QueryTrace child : children) {
+            preAggregationSampleSize += child.getPreAggregationSampleSize();
+            numSeries += child.getNumSeries();
+        }
+    }
 
     public static QueryTrace of(final Identifier what) {
         return new QueryTrace(what, 0L, ImmutableList.of());
@@ -49,6 +62,15 @@ public class QueryTrace {
         final Identifier what, final long elapsed, final List<QueryTrace> children
     ) {
         return new QueryTrace(what, elapsed, children);
+    }
+
+    public static QueryTrace of(
+        final Identifier what, final long elapsed, final List<QueryTrace> children,
+        final long preAggregationSampleSize, final long numSeries
+    ) {
+        QueryTrace qt = new QueryTrace(what, elapsed, children, preAggregationSampleSize,
+                                       numSeries);
+        return qt;
     }
 
     public void formatTrace(PrintWriter out) {
--- a/heroic-core/src/main/java/com/spotify/heroic/CoreQueryManager.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/CoreQueryManager.java
@@ -97,12 +97,14 @@ public class CoreQueryManager implements QueryManager {
     private final QueryCache queryCache;
     private final AggregationFactory aggregations;
     private final OptionalLimit groupLimit;
+    private final QueryLogger queryLogger;
 
     @Inject
     public CoreQueryManager(
         @Named("features") final Features features, final AsyncFramework async,
         final ClusterManager cluster, final QueryParser parser, final QueryCache queryCache,
-        final AggregationFactory aggregations, @Named("groupLimit") final OptionalLimit groupLimit
+        final AggregationFactory aggregations, @Named("groupLimit") final OptionalLimit groupLimit,
+        CoreQueryLogger queryLogger
     ) {
         this.features = features;
         this.async = async;
@@ -111,6 +113,7 @@ public class CoreQueryManager implements QueryManager {
         this.queryCache = queryCache;
         this.aggregations = aggregations;
         this.groupLimit = groupLimit;
+        this.queryLogger = queryLogger;
     }
 
     @Override
@@ -176,6 +179,11 @@ public class CoreQueryManager implements QueryManager {
 
         @Override
         public AsyncFuture<QueryResult> query(Query q) {
+            boolean shouldLogQueries = features.hasFeature(Feature.LOG_QUERIES);
+            if (shouldLogQueries) {
+                queryLogger.logQueryAccess(q);
+            }
+
             final List<AsyncFuture<QueryResultPart>> futures = new ArrayList<>();
 
             final MetricType source = q.getSource().orElse(MetricType.POINT);
@@ -232,8 +240,17 @@ public class CoreQueryManager implements QueryManager {
 
                 final OptionalLimit limit = options.getGroupLimit().orElse(groupLimit);
 
-                return async.collect(futures,
+                AsyncFuture<QueryResult> queryResult = async.collect(futures,
                     QueryResult.collectParts(QUERY, range, combiner, limit));
+
+                if (shouldLogQueries) {
+                    queryResult = queryResult.directTransform(_queryResult -> {
+                        queryLogger.logQueryResolved(q, _queryResult);
+                        return _queryResult;
+                    });
+                }
+
+                return queryResult;
             });
         }
 
--- a/heroic-core/src/main/java/com/spotify/heroic/http/query/QueryResource.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/http/query/QueryResource.java
@@ -21,6 +21,8 @@
 
 package com.spotify.heroic.http.query;
 
+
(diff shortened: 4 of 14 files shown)


heroic-component/src/main/java/com/spotify/heroic/metric/QueryResult.java
```diff
@@ -95,8 +114,51 @@
...
+                                                      trace, new ResultLimits(limits.build()));
+
+            queryResult.logQueryInfo(request, requestMetadata,
+                                     logQueries, logQueriesThresholdDataPoints);
+
+            return queryResult;
         };
     }
+
+    public void logQueryInfo(final FullQuery.Request request,
```
udoprog: This does not make sense to live in `QueryResult`. Move it to an interface (e.g. `QueryLogger`) and provide an implementation to it in core. Then make use of it in CoreQueryManager.

Similarly do not use static fields for `totalQueriesProcessed` and `queriesAboveThreshold`. When they live outside of `QueryResult`, this is not a problem.

Similarly for `queryLog`, which would be a member of the `QueryLogger` implementation.

heroic-core/src/main/java/com/spotify/heroic/http/query/QueryHttpRequestMetadata.java
```diff
@@ -0,0 +1,44 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.heroic.http.query;
+
+import com.spotify.heroic.QueryRequestMetadata;
+import javax.servlet.http.HttpServletRequest;
+
+public class QueryHttpRequestMetadata extends QueryRequestMetadata {
```
udoprog: There is nothing special about `QueryHttpRequestMetadata` that is not already covered by `QueryRequestMetadata`, similarly the factory method `of(HttpServletRequest)` should probably live in the http package and not here.

heroic-core/src/main/java/com/spotify/heroic/metric/MetricManagerModule.java
```diff
@@ -88,6 +90,16 @@
      */
     private final boolean failOnLimits;
 
+    /**
+     * If {@code true}, will enable logging of queries
+     */
+    private final boolean logQueries;
```
udoprog: This should be a `Feature` flag. It's already supported in the query path. (lookup the `Feature` class and its usages).

heroic-component/src/main/java/com/spotify/heroic/metric/QueryResult.java
```diff
@@ -95,8 +114,51 @@
...
+        }
+
+        totalQueriesProcessed.increment();
+
+        long totDataPoints = 0;
+        for (ShardedResultGroup g : groups) {
+            totDataPoints += g.getMetrics().getData().size();
+        }
+
+        if (totDataPoints < logQueriesThresholdDataPoints.orElse(
```
udoprog:
```java
if (logQueriesThresholdDataPoints.isGreaterOrEqual(totalDataPoints)) {
    /*  */
}
```

heroic-component/src/main/java/com/spotify/heroic/metric/QueryResult.java
```diff
@@ -95,8 +114,51 @@
...
+    ) {
+        if (!logQueries) {
+            return;
+        }
+
+        totalQueriesProcessed.increment();
+
+        long totDataPoints = 0;
+        for (ShardedResultGroup g : groups) {
+            totDataPoints += g.getMetrics().getData().size();
```
udoprog: This does not account for how much data was being processed, only how much data is present post-aggregation.

heroic-core/src/main/java/com/spotify/heroic/CoreQueryManager.java
```diff
@@ -97,12 +97,16 @@
     private final QueryCache queryCache;
     private final AggregationFactory aggregations;
     private final OptionalLimit groupLimit;
+    private boolean logQueries;
+    private OptionalLimit logQueriesThresholdDataPoints;
```
udoprog: The threshold should probably be time-based instead of size.

heroic-component/src/main/java/com/spotify/heroic/QueryRequestMetadata.java
```diff
@@ -0,0 +1,49 @@
...
+import lombok.Data;
+
+@Data
+public class QueryRequestMetadata {
+    private final String  remoteAddr;
+    private final String  remoteHost;
+    private final int     remotePort;
+    private final String  remoteUserAgent;
+
+    public QueryRequestMetadata() {
```
udoprog: Use a factory method for this.

```
public class QueryRequestMetadata {
    public static QueryRequestMetadata empty() {
        return new QueryRequestMetadata(...);
    }
}
```

heroic-component/src/main/java/com/spotify/heroic/QueryRequestMetadata.java
```diff
@@ -0,0 +1,49 @@
...
+        remoteAddr = "";
+        remoteHost = "";
+        remotePort = 0;
+        remoteUserAgent = "";
+    }
+
+    public QueryRequestMetadata(final String  remoteAddr,
```
udoprog: Lombok's `@Data` already generates this.

heroic-component/src/main/java/com/spotify/heroic/metric/QueryResult.java
```diff
@@ -95,8 +114,51 @@
...
+
+        if (totDataPoints < logQueriesThresholdDataPoints.orElse(
+                                    OptionalLimit.of(0)).asLong().get()) {
+            return;
+        }
+
+        long currQueriesAboveThreshold = queriesAboveThreshold.incrementAndGet();
+
+        boolean isIPv6 = requestMetadata.getRemoteAddr().indexOf(':') != -1;
+        queryLog.debug("[Query] (" + currQueriesAboveThreshold + "/" + totalQueriesProcessed + ")" +
```
udoprog: Also, use `trace` log level.

Don't align statements like this. Use the format generated by the provided [code-style](https://github.com/spotify/heroic/tree/master/idea) or escape it using `@formatter:off` comments as described in the [README](https://github.com/spotify/heroic#bypassing-validation).

Special formatting (without escaping) causes future issues when someone tries to build on top of your change. And they accidentally format your code, making their change harder to decipher.

heroic-component/src/main/java/com/spotify/heroic/Query.java
```diff
@@ -38,4 +38,6 @@
     private final Optional<QueryOptions> options;
     /* set of experimental features to enable */
     private final Optional<FeatureSet> features;
+    private Optional<QueryRequestMetadata> requestMetadata =
```
udoprog: Make `final`, and don't assign a default value. This will default to `Optional.empty()` if it is not specified.

heroic-core/src/main/java/com/spotify/heroic/http/query/QueryHttpRequestMetadata.java
gabrielgerhardsson: I did this since HttpServer is in heroic-core but QueryResponse is in heroic-component, and I didn't want to add dependency on javax.servlet.http.HttpServletRequest in heroic-component. However, your above comment about moving the main query logging out of QueryResponse is gonna make that a moot point I guess :)

gabrielgerhardsson:
Thanks for the great code review! I'll get cranking.

udoprog:
I think this relates to #126 somehow, @juruen?

heroic-core/src/main/java/com/spotify/heroic/CoreQueryManager.java
gabrielgerhardsson: Touched on this subject earlier today with @juruen. He commented that he had seen that when a big query hit the system it would also slow down other queries, so we'd perhaps see a large swath of queries being logged at that time. I'll add time based threshold tho.

heroic-component/src/main/java/com/spotify/heroic/metric/QueryResult.java
gabrielgerhardsson: I'll clarify that, so that it says "postAggregation..." instead of totDataPoints. 

Also, we were experimenting with taking AggregationResult -> getResults -> getCounters -> get("Aggregation.sampleSize"), and logging that value as preAggregationSampleSize (see the changes in LocalMetricManager and QueryLogger). Would you say that's correct, or is there a better way of getting to know how much data we had before aggregation?

heroic-core/src/main/java/com/spotify/heroic/metric/MetricManagerModule.java
gabrielgerhardsson: I've done this now, but am wondering how this would be configured to be enabled by default without having to recompile.
I know that Feature flags can be set in the query - can a default also be set in the configuration file?

gabrielgerhardsson:
@udoprog 
This is now ready for review again.

I force pushed a large update to this PR, which unfortunately removed the previous code review comments. 
I can say tho that I went through all comments and took action for each one. Hopefully it's better now.

One question from my side: Is it possible to enable a Feature flag in the configuration file? We'd like to be able to enable the new Feature flag com.spotify.heroic.log_queries for all queries (that doesn't specify otherwise). It can be enabled by default in code, but that's not optimal.

heroic-core/src/main/java/com/spotify/heroic/http/query/QueryResource.java
```diff
@@ -143,6 +177,28 @@ private void bindMetricsResponse(
                 r.getLimits()));
     }
 
+    private QueryMetrics parseQueryMetricsFrom(String queryString) {
+        QueryMetrics query = null;
+        try {
+            query = objectMapper.readValue(queryString, QueryMetrics.class);
+        } catch (IOException e) {
+            log.info("Failed to deserialize JSON to QueryMetrics object! \"" + queryString + "\"");
+            log.info("Error: " + e.toString());
```
udoprog: Logging is the wrong choice, just throw the exception raised here and the user should be notified.

I'm a bit suspicious of this approach that it would maintain all user interaction (specifically supply the user with correctly formatted `4xx` responses). You are probably better of serializing the `QueryMetrics` instead.

heroic-component/src/main/java/com/spotify/heroic/metric/QueryTrace.java
```diff
@@ -25,18 +25,31 @@
...
 import java.util.concurrent.TimeUnit;
 
+@AllArgsConstructor
 @Data
 public class QueryTrace {
     private final Identifier what;
     private final long elapsed;
     private final List<QueryTrace> children;
 
+    private long preAggregationSampleSize;
```
udoprog: Everything should be `final`, all data structures are immutable.

heroic-core/src/main/java/com/spotify/heroic/CoreQueryManager.java
```diff
@@ -232,8 +240,17 @@ public Aggregation visitString(final StringExpression e) {
 
                 final OptionalLimit limit = options.getGroupLimit().orElse(groupLimit);
 
-                return async.collect(futures,
+                AsyncFuture<QueryResult> queryResult = async.collect(futures,
                     QueryResult.collectParts(QUERY, range, combiner, limit));
+
+                if (shouldLogQueries) {
+                    queryResult = queryResult.directTransform(_queryResult -> {
```
udoprog: We generally don't use any underscore convention.

Also, no need to transform the result, do this instead:
```java
queryResult.onResolved(result -> {
  queryLogger.logQueryResolved(q, result);
});
```

heroic-core/src/main/java/com/spotify/heroic/metric/MetricManagerModule.java
```diff
@@ -88,6 +89,11 @@
      */
     private final boolean failOnLimits;
 
+    /**
+     * Limit which queries that are logged
+     */
+    private final OptionalLimit logQueriesThresholdDataPoints;
```
udoprog: This variable name reads a bit weird, either `DataPoints` should come before `Threshold`, or use something simpler like `logQueryThreshold`.

heroic-core/src/main/java/com/spotify/heroic/http/query/QueryResource.java
```diff
@@ -143,6 +177,28 @@ private void bindMetricsResponse(
...
+        return query;
+    }
+
+    private QueryBatch parseQueryBatchFrom(String queryString) {
+        QueryBatch query = null;
+        try {
+            query = objectMapper.readValue(queryString, QueryBatch.class);
+        } catch (IOException e) {
+            log.info("Failed to deserialize JSON to QueryBatch object! \"" + queryString + "\"");
+            log.info("Error: " + e.toString());
```
udoprog: Same here.

heroic-component/src/main/java/com/spotify/heroic/QueryOriginContext.java
```diff
@@ -0,0 +1,52 @@
...
+    private final String remoteUserAgent;
+    private final String remoteClientId;
+    private final UUID   queryId;
+    private final String queryString;
+
+    public static QueryOriginContext empty() {
+        return new QueryOriginContext("", "", 0, "", "", UUID.randomUUID(), "");
+    }
+
+    public static QueryOriginContext of(final String  remoteAddr,
```
udoprog: Don't align parameters. Also, the correct way to align this is a line break before the opening and closing paranthesis.

Are you using the provided formatter?

heroic-core/src/main/java/com/spotify/heroic/metric/MetricManagerModule.java
```diff
@@ -54,6 +54,7 @@
 public class MetricManagerModule {
     public static final int DEFAULT_FETCH_PARALLELISM = 100;
     public static final boolean DEFAULT_FAIL_ON_LIMITS = false;
+    public static final long DEFAULT_LOG_QUERIES_THRESHOLD = 1000000;
```
udoprog: Make the type of this `OptionalLimit`.

PR #136, opened 2016-11-22
onComplete log severity "DEBUG"

mykolasmith:
Having a WARN message every time the BigTable client completes a transaction results in extremely verbose logging for high throughput applications.


juruen:
I believe that's not the BT client. I was under the impression that's the HTTP client that triggers the request.

@udoprog should be able to clarify this.

mykolasmith:
@juruen I assumed it was the HTTP client triggering a call to the BigTable client since every time I hit `/write`, I get the following log message:

```2016-11-22 21:07:32,512 WARN c.s.h.c.CoreJavaxRestFramework [bigtable-batch-pool-287] Client completed```

Could this actually be slowing down the HTTP client? I imagine it has to finish logging that message before it can move on and process the next request (purely speculating: I don't know how the JavaxRestFramework actually works)

udoprog:
Looks like a good change. I even think `debug` might be appropriate in this case.

> Could this actually be slowing down the HTTP client? I imagine it has to finish logging that message before it can move on and process the next request (purely speculating: I don't know how the JavaxRestFramework actually works)

There shouldn't be any client slowdown. The Jersey server is configured to be used asynchronously which causes it to have distinct thread pools for accepting requests, and passing them into the handler.
This even seems to be utilizing cycles from the pool that finished the request, rather than any of those internal ones (`bigtable-batch-pool-287`).

I wouldn't rule it out completely without a benchmark though, but so far I've seen no proof that the HTTP layer can't keep up with the rest of the stack.

mykolasmith:
@udoprog Updated to be debug instead of info. Please merge?

udoprog:
@mykolasmith Sorry, I've had a few things occupying me this week. Could you fix the commit message to reflect the corrected log level?

mykolasmith:
Updated commit message and pull request to reflect debug log level.

udoprog:
@mykolasmith looks like you got a few more commits than you bargained for.

mykolasmith:
@udoprog Sorry, I clicked "Update branch". Going forward, I will not do that. I rolled back those merges. It should be just my change now.

udoprog:
If you look in the list of commits for this PR you can still see 5 commits.

PR #137, opened 2016-11-29
[aggregation] PointsAbove and PointsBelow

mykolasmith:
The `pointsabove` and `pointsbelow` aggregations work similar to the `abovek` and `belowk` aggregations, except they act on a per-point basis, rather than a per-series basis. The use case we have for this is a follows:

- A metric comes from our devices, incrementing over time and occasionally resetting.
- We want to sum all of the positive deltas (effectively ignoring the resets), so as to get the total "forward progress" of the metric over a given time period.

Our aggregations look something like this (tested on real data in our QA BigTable cluster and confirmed working):

```
"aggregation": {
	"type": "chain",
	"chain": [
		{"type": "delta"},
		{"type": "pointsabove", "k": 0},
		{"type": "sum", "sampling": {"size": "1h"}}
	]
}
```

### DIFF
--- a/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/FilterPointsThresholdStrategy.java
+++ b/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/FilterPointsThresholdStrategy.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright (c) 2015 Spotify AB.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.heroic.aggregation.simple;
+
+import com.spotify.heroic.metric.MetricCollection;
+import com.spotify.heroic.metric.MetricType;
+import com.spotify.heroic.metric.Point;
+import lombok.Data;
+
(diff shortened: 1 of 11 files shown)


udoprog:
LGTM, all though I would drop the `K` suffix in the class names as well and call the parameter `value`, not `k`.

In hindsight the use of `k` instead of descriptive parameters (e.g. `count`/`value`) is confusing.

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/FilterPointsKThresholdStrategy.java
```diff
@@ -0,0 +1,47 @@
...
+@Data
+public class FilterPointsKThresholdStrategy implements MetricMappingStrategy{
+    private final FilterKThresholdType filterType;
+    private final double k;
+
+    @Override
+    public MetricCollection apply(MetricCollection metrics) {
+        return MetricCollection.build(
+            MetricType.POINT,
+            filterWithThreshold(metrics.getDataAs(Point.class))
```
udoprog: This would cause a runtime exception if the data type is not `Point`. Better make sure that all the available types can be handled or that the aggregation ignores non-point data.

mykolasmith:
@denissytenkov can you please make these changes?

1. Change class names to not include "k"
2. Change parameter name from "k" to "value"
3. Ensure that the aggregation ignores non-point data.

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/FilterPointsKThresholdStrategy.java
mykolasmith: Would something like this work?

```
    public MetricCollection apply(MetricCollection metrics) {
        if (metrics.getType() == MetricType.POINT) {
            return MetricCollection.build(
                MetricType.POINT,
                filterWithThreshold(metrics.getDataAs(Point.class))
            );
        } else {
            return metrics;
        }
    }
```

udoprog: Yeah, that should work.

mykolasmith:
@udoprog fixed those three things...

udoprog:
Restarted the build, maybe Travis will do us the honor this time...

mykolasmith:
@udoprog Checks passed!

udoprog:
Waiting for @parmus to do the final look over and merge.

hampusohlsson:
@parmus status on this PR?

parmus:
@hampusohlsson Sorry about the delay.

PR #141, opened 2016-11-30
Set Bigtable batchSize as a parameter

zfrank:
Allow a new setting for bigtable metric backend: batchSize.
This will set the size of BulkMutations when initializing BigTable.

### DIFF
--- a/metric/bigtable/src/main/java/com/spotify/heroic/analytics/bigtable/BigtableAnalyticsModule.java
+++ b/metric/bigtable/src/main/java/com/spotify/heroic/analytics/bigtable/BigtableAnalyticsModule.java
@@ -49,6 +49,7 @@ import java.util.Optional;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 
+
 @ToString
 @RequiredArgsConstructor
 @Module
@@ -93,7 +94,7 @@ public class BigtableAnalyticsModule implements AnalyticsModule {
                 return async.call(
                     new BigtableConnectionBuilder(project, cluster, credentials, async,
                         executorService, DEFAULT_DISABLE_BULK_MUTATIONS,
-                        DEFAULT_FLUSH_INTERVAL_SECONDS));
+                        DEFAULT_FLUSH_INTERVAL_SECONDS, Optional.empty()));
             }
 
             @Override
--- a/metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableConnectionBuilder.java
+++ b/metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableConnectionBuilder.java
@@ -23,6 +23,7 @@ package com.spotify.heroic.metric.bigtable;
 
 import com.google.appengine.repackaged.com.google.common.collect.ImmutableList;
 import com.google.cloud.bigtable.config.BigtableOptions;
+import com.google.cloud.bigtable.config.BulkOptions;
 import com.google.cloud.bigtable.config.CredentialOptions;
 import com.google.cloud.bigtable.config.RetryOptions;
 import com.google.cloud.bigtable.grpc.BigtableSession;
@@ -38,6 +39,7 @@ import eu.toolchain.async.AsyncFuture;
 import lombok.RequiredArgsConstructor;
 import lombok.ToString;
 
+import java.util.Optional;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ExecutorService;
 
@@ -56,6 +58,7 @@ public class BigtableConnectionBuilder implements Callable<BigtableConnection> {
 
     private final boolean disableBulkMutations;
     private final int flushIntervalSeconds;
+    private final Optional<Integer> batchSize;
 
     @Override
     public BigtableConnection call() throws Exception {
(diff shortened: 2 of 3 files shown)


metric/bigtable/src/main/java/com/spotify/heroic/analytics/bigtable/BigtableAnalyticsModule.java
```diff
@@ -87,13 +91,14 @@ public AnalyticsReporter reporter(final HeroicReporter reporter) {
     public Managed<BigtableConnection> connection(
         final AsyncFramework async, final ExecutorService executorService
     ) {
+        Optional<Integer> batchSize = empty();
```
udoprog: Inline this variable, or declare as a default constant to follow the local convention?

metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableConnectionBuilder.java
```diff
@@ -66,6 +69,10 @@ public BigtableConnection call() throws Exception {
             .setAllowRetriesWithoutTimestamp(true)
             .build();
 
+        final BulkOptions bulkOptions;
+        bulkOptions = batchSize
```
udoprog: No need for a separate declaration.

zfrank: Done!

metric/bigtable/src/main/java/com/spotify/heroic/analytics/bigtable/BigtableAnalyticsModule.java
zfrank: The issue is that we would like the default to be what Google has set in the client library, since one would assume they know better and they would change defaults for a good reason.

We can do this by either initializing the vaule to BulkOptions.BIGTABLE_BULK_MAX_ROW_KEY_COUNT_DEFAULT or not initializing it at all so that the default gets applied instead. I went for the former because I fear the latter will create a dependency between both modules and potentially make the code harder to read/maintain in the long term.

What do you suggest?

udoprog: This was only about inlining the variable, instead of declaring it separately from where it is being used:

```java
Optional<Integer> batchSize = empty();
/* ... */
new BigtableConnectionBuilder(/*  */, batchSize)
```

Into:

```java
new BigtableConnectionBuilder(/*  */, empty())
```

I think you are already falling back to the default value (without accessing the constant) by using the the builder, so that should not be a problem?

Speaking of which, the static import for `BIGTABLE_BULK_MAX_ROW_KEY_COUNT_DEFAULT` never seems to be used?
It is also worth keeping in mind that `javac` will inline constants, so it is recommended against sharing them across libraries:
http://stackoverflow.com/questions/3524150/is-it-possible-to-disable-javacs-inlining-of-static-final-variables

zfrank: Sorry, I misunderstood you.
When you suggested 

> or declare as a default constant to follow the local convention?

 I thought you preferred that we assigned a default value to batchSize at Bigtable*Module creation time, and that is precisely what we would like to avoid so that Google's defaults get applied instead. I completely missed the inlining part. I'm adding a new commit for that now.

Also, good catch with the leftover `BIGTABLE_BULK_MAX_ROW_KEY_COUNT_DEFAULT` imports. That is precisely what I wanted to avoid :)

metric/bigtable/src/main/java/com/spotify/heroic/analytics/bigtable/BigtableAnalyticsModule.java
```diff
@@ -143,10 +146,10 @@ public static Builder builder() {
...
     @NoArgsConstructor
     public static class Builder implements AnalyticsModule.Builder {
-        private Optional<String> project = Optional.empty();
-        private Optional<String> instance = Optional.empty();
-        private Optional<CredentialsBuilder> credentials = Optional.empty();
-        private Optional<Integer> maxPendingReports = Optional.empty();
+        private Optional<String> project = empty();
+        private Optional<String> instance = empty();
+        private Optional<CredentialsBuilder> credentials = empty();
+        private Optional<Integer> maxPendingReports = empty();
```
udoprog: Keep in mind that, these kind of refactoring tend to make things harder to merge when there are conflicts. Try to avoid them in the future, or do them in a distinct commit.

metric/bigtable/src/main/java/com/spotify/heroic/analytics/bigtable/BigtableAnalyticsModule.java
```diff
@@ -62,6 +64,7 @@
     public static final boolean DEFAULT_DISABLE_BULK_MUTATIONS = false;
     public static final int DEFAULT_FLUSH_INTERVAL_SECONDS = 2;
 
+
```
udoprog: Extra line for not apparent reason?

udoprog:
Squash the commits and then I'm happy.

metric/bigtable/src/main/java/com/spotify/heroic/analytics/bigtable/BigtableAnalyticsModule.java
zfrank: That was done automatically by my IDE and I didn't think it could cause any harm. I revered it.

metric/bigtable/src/main/java/com/spotify/heroic/analytics/bigtable/BigtableAnalyticsModule.java
zfrank: And artifact from previous edits. I'm removing it.

PR #143, opened 2016-12-01
rename Client-Id header to make it more generic

zfrank:
The current Client-Id header has a typo and points to a specfic client
implementation. It would be more useful to keep it generic.


zfrank:
This change clashes with PR #144 so I'm closing it.

PR #144, opened 2016-12-01
Log incoming query before processing it

gabrielgerhardsson:
We already had a query.log, it was renamed to query.done.log
Now there's also a query.access.log where incoming queries are printed, together with a query-unique UUID so that query.access.log and query.done.log can be matched with each other. If a node dies, it should now be possible to know which queries that were active/outstanding when it happened.

log4j details:
* Change: query.log renamed to query.done.log
* New Logger: query.access.log
* Log level changed to TRACE for query logs

### DIFF
--- a/heroic-core/src/main/java/com/spotify/heroic/http/query/CoreQueryOriginContextFactory.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/http/query/CoreQueryOriginContextFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2015 Spotify AB.
+ * Copyright (c) 2016 Spotify AB.
  *
  * Licensed to the Apache Software Foundation (ASF) under one
  * or more contributor license agreements.  See the NOTICE file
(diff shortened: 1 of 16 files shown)


heroic-core/src/main/java/com/spotify/heroic/http/query/CoreQueryOriginContextFactory.java
```diff
@@ -21,19 +21,20 @@
...
+import com.spotify.heroic.QueryOriginContext;
 import javax.servlet.http.HttpServletRequest;
 
-public class CoreQueryRequestMetadataFactory {
+public class CoreQueryOriginContextFactory {
 
-    public static QueryRequestMetadata create(HttpServletRequest httpServletRequest) {
+    public static QueryOriginContext create(HttpServletRequest httpServletRequest, String query) {
         String userAgent = httpServletRequest.getHeader("User-Agent");
         String clientId = httpServletRequest.getHeader("X-Alient-Client-Id");
```
zfrank: Could we changed the header name to "X-Client-Id" instead?

gabrielgerhardsson: Absolutely

PR #147, opened 2016-12-02
[travis,pom] tweak build configuration to optimize thread usage

udoprog:
Travis only provides two threads, this attempts to make sure we only use two threads and retain sensible defaults when building outside of Travis.


udoprog:
Travis is hung, built tests locally by me, so forcing the PR.

Issue #150, opened 2016-12-08
Group aggregation using tags and key?


hampusohlsson:
Running into an issue similar to spotify/heroic#138 where I need to group by tags **and** the timeseries key.

In our setup, the key is what differentiates the time series (such as the id of a device). Every device have different metrics stored as tags. These metric names may or may not be the same across devices (temperature for example).

Right now if I query the temperature from two different keys (series), they are merged into one result set. Is there any way get around that behavior?

udoprog:
Hey!

At the moment you should consider using tags to structure dimensional things since keys can only be filtered. It is possible to extend the grouping aggregation to support keys, but I can't promise it will be done anytime soon.

As a hint to anyone wanting to give it ago, it would involve changing the implementation of the key method in grouping aggregation*, and add a `Optional<Boolean> byKey` flag to the model** and instance*** classes.

*: https://github.com/spotify/heroic/blob/master/heroic-component/src/main/java/com/spotify/heroic/aggregation/GroupingAggregation.java#L65
**: https://github.com/spotify/heroic/blob/master/heroic-component/src/main/java/com/spotify/heroic/aggregation/Group.java#L37
***: https://github.com/spotify/heroic/blob/master/heroic-component/src/main/java/com/spotify/heroic/aggregation/GroupingAggregation.java#L51

hampusohlsson:
Ok thanks for the information. Will probably do explicit query batching for now as a temporary solution.

(Btw the docs are missing the `queries` parameter needed for batching, took me a while to figure out)

```
{
    "a": { ... }
    "b": { ... }
}
```
should be
```
{
    "queries": {
        "a": { ... }
        "b": { ... }
    }
}
```

udoprog:
Thanks for the tip.

hexedpackets:
Closing due to age, feel free to re-open if this is still an issue.

PR #151, opened 2016-12-09
[docs] Added missing queries parameter to batch endpoint

hampusohlsson:
Related to spotify/heroic#150


udoprog:
Great stuff, thanks!

PR #156, opened 2016-12-13
[aggregations] Filter out time series with 0 datapoints in TopK/BottomK

parmus:
Metrics with 0 data points (e.g. time series, which are no longer reported but still in the index) should not be considered for TopK and BottomK.


aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/FilterKAreaStrategy.java
```diff
@@ -37,6 +37,7 @@
     public <T> List<T> filter(List<FilterableMetrics<T>> metrics) {
         return metrics
             .stream()
+            .filter(m -> m.getMetricSupplier().get().size() > 0)
```
gabrielgerhardsson: Please add comment describing what this filter does

PR #162, opened 2016-12-14
[component, core, statistics] Add counters for tracking samples in mem

gabrielgerhardsson:
New counters:
1) sample-size-live:
Incremented on data being read from backend.
Decremented when processing of data is done, i.e. when it is expected that data is freed.
2) sample-size-accumulated:
Incremented on data being read from backend.
Decremented when the whole request finishes.

Also started adding test scaffolding for LocalMetricManagerTest. It's too much work to do the full test at this time, but here's at least the test scaffolding.


udoprog:
Build fails.

udoprog:
Bypassing status checks since Travis seems unhappy right now.

PR #168, opened 2016-12-15
[core] Faster shutdown of Heroic

gabrielgerhardsson:
This change affects the shutdown behaviour of core ScheduledExecutorService so that it doesn't wait for scheduled/delayed tasks that aren't yet running but are scheduled to run at some time in the future. Previous to this the cluster refresh in clustered configurations would always be waited upon, introducing a delay in the shutdown.


udoprog:
Just to note: build time goes from 12:50 to 6:50 because if this!

Issue #172, opened 2016-12-20
More detailed tracing configuration


udoprog:
Currently there is only one level of tracing which is overly detailed for common queries. Tracing should instead be configurable with different levels that determines how much tracing information is made available.

#### Suggestion

This suggests the addition of tracing levels. The levels suggested initially are the following:

* `TracingLevel.NONE`
* `TracingLevel.DEFAULT`
* `TracingLevel.ALL`

Note that these only applies when tracing has actually been enabled.

On top of this, the API will be extend to support the following values for the `tracing` option.

```javascript
// NONE
{ "tracing": false }
// DEFAULT
{ "tracing": true }
// ALL
{ "tracing": "all" }
```

In order to determine if a specific level is enabled or not, the `Tracing` and `QueryTrace` API will be extended with the following methods:

```java
interface Tracing {
    /**
     * Check if tracing is enabled for the given level or not.
     *
     * @return {@code true} if tracing is enabled, {@code false} otherwise
     */
    boolean isEnabled(TracingLevel level);

    /**
     * Create a new watch if the current level is equal to or larger than the given level.
     *
     * The level restriction will be propagated through all the child traces.
     *
     * @param what what is being watched
     * @param level level to match against
     * @return a {@link com.spotify.heroic.metric.QueryTrace.NamedWatch}
     */
    QueryTrace.NamedWatch watch(QueryTrace.Identifier what, TracingLevel level); 
}
```

```java
interface QueryTrace {
    interface NamedWatch {
        /**
         * Create another watch with the same configuration as this watch (active or passive)  if the current level is equal to or larger than the given level.
         *
         * @param what what is being watched
         * @param level level to match against
         * @return a {@link com.spotify.heroic.metric.QueryTrace.NamedWatch}
         */
        NamedWatch watch(Identifier what, TracingLevel level);
    }
}
```

A `NamedWatch` will be passive if the given level does not match the one which was configured with the `Tracing` option. This means that the an entire sub-hierarchy will be ignored.

PR #173, opened 2016-12-20
Make use of improved tracing framework and deprecate old one

udoprog:
This removes use of deprecated APIs related to tracing in favor of the new one.

Currently tracing will be <em>disabled</em> by default when this is merged, this is because it has become a fair bit more granular than before. It might be desirable to incorporate more limited tracing by default in the future by extending how it can be configured (see #172).


udoprog:
Closing this in favor of incrementally adding it. But the branch this is based off can be used as a basis.

PR #176, opened 2016-12-29
[all] Use Tracing type in QueryOptions

udoprog:


gabrielgerhardsson:
@udoprog: I believe this commit broke AbstractClusterQueryIT::distributedQueryTraceTest(), since QueryTrace is now disabled by default.

PR #182, opened 2017-01-04
[all] Add metric for ratio of live Kafka ConsumerThread's

gabrielgerhardsson:
The new metric is of type "ratio" (added in this commit), and named
"consumer-threads-live-ratio". In normal operation it should always be 1.
Anything below 1 (except during startup), is likely a good basis for signalling
an alert.

### DIFF
--- a/consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/ConsumerThread.java
+++ b/consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/ConsumerThread.java
@@ -27,16 +27,15 @@ import com.spotify.heroic.statistics.ConsumerReporter;
 import eu.toolchain.async.AsyncFramework;
 import eu.toolchain.async.AsyncFuture;
 import eu.toolchain.async.ResolvableFuture;
-import kafka.consumer.KafkaStream;
-import kafka.message.MessageAndMetadata;
-import lombok.extern.slf4j.Slf4j;
-
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.atomic.LongAdder;
+import kafka.consumer.KafkaStream;
+import kafka.message.MessageAndMetadata;
+import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public final class ConsumerThread extends Thread {
@@ -83,21 +82,31 @@ public final class ConsumerThread extends Thread {
     public void run() {
         log.info("{}: Starting thread", name);
 
-        active.incrementAndGet();
+        threadIsStarting();
 
         try {
             guardedRun();
         } catch (final Throwable e) {
             log.error("{}: Error in thread", name, e);
-            active.decrementAndGet();
+            threadIsStopping();
             stopFuture.fail(e);
             return;
         }
 
         log.info("{}: Stopping thread", name);
-        active.decrementAndGet();
+        threadIsStopping();
         stopFuture.resolve(null);
         return;
+    }
+
+    private void threadIsStarting() {
+        active.incrementAndGet();
+        reporter.reportConsumerThreadsIncrement();
+    }
+
+    private void threadIsStopping() {
+        active.decrementAndGet();
+        reporter.reportConsumerThreadsDecrement();
     }
 
     public AsyncFuture<Void> pauseConsumption() {
--- a/statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticConsumerReporter.java
+++ b/statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticConsumerReporter.java
@@ -40,6 +40,7 @@ public class SemanticConsumerReporter implements ConsumerReporter {
     private final Meter messageIn;
     private final Meter messageError;
     private final Meter consumerSchemaError;
+    private final SemanticRatioGauge consumerThreadsLiveRatio;
     private final Histogram messageSize;
     private final Histogram messageDrift;
 
@@ -52,6 +53,9 @@ public class SemanticConsumerReporter implements ConsumerReporter {
         messageError = registry.meter(base.tagged("what", "message-error", "unit", Units.FAILURE));
         consumerSchemaError =
             registry.meter(base.tagged("what", "consumer-schema-error", "unit", Units.FAILURE));
+        consumerThreadsLiveRatio = new SemanticRatioGauge();
+        registry.register(base.tagged("what", "consumer-threads-live-ratio", "unit", Units.RATIO),
+            consumerThreadsLiveRatio);
         messageSize = registry.histogram(base.tagged("what", "message-size", "unit", Units.BYTE));
         messageDrift =
             registry.histogram(base.tagged("what", "message-drift", "unit", Units.MILLISECOND));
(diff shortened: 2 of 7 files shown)


statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticConsumerReporter.java
```diff
@@ -40,6 +41,8 @@
     private final Meter messageIn;
     private final Meter messageError;
     private final Meter consumerSchemaError;
+    private Optional<Long> consumerThreadsWanted;
```
udoprog: This needs to be synchronized, like an `AtomicLong`. Otherwise there are no guarantees updates will be visible across threads.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticRatioGauge.java
```diff
@@ -0,0 +1,41 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.heroic.statistics.semantic;
+
+import com.codahale.metrics.RatioGauge;
+
+public class SemanticRatioGauge extends RatioGauge {
```
udoprog: The typical way that you build ratio gauge is that they are based of two other structures which measure things:
See http://metrics.dropwizard.io/3.1.0/manual/core/

This way, it doesn't hold any of its own internal state.

consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/ConsumerThread.java
```diff
@@ -83,23 +82,38 @@ public ConsumerThread(
...
         log.info("{}: Stopping thread", name);
-        active.decrementAndGet();
+        threadIsStopping();
         stopFuture.resolve(null);
         return;
     }
 
+    private void threadIsStarting() {
+        // Increment counter and report atomically
+        synchronized (active) {
```
udoprog: you can avoid synchronizing by making the interaction with the reporting framework:
```java
reporter.incrementActiveThreads()
/*  */
reporter.decrementActiveThreads()
```

And have the `active-threads` just being a counter.

Because the reporter is shared across threads, all synchronization has to happen there. There are no guarantees an arbitrary caller will yield to this lock.

I also realize it seems compelling to have the `active` counter being in-sync between what's being reported and what is not. They are used for two different things, so it's not a big deal if they are not.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticConsumerReporter.java
gabrielgerhardsson: It's only set once, before the ConsumerThread threads are started. Tho as you note below, we don't know if the callee will change in the future so the correct way is to ensure it's thread safe. Thanks for the pointer.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticRatioGauge.java
gabrielgerhardsson: Ah I see. 
The most simple implementation would then have SemanticRatioGauge contain two Double's, and do "return Ratio.of(...)" in getRatio(). Would you agree? Or did you mean that we should have a "Metric"-extending structure inside a RatioGauge?

consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/ConsumerThread.java
gabrielgerhardsson: Very true, thanks for the pointer. Of course the synchronisation should be around the data and not at the caller. 
As you noted, I was coming from a mindset of keeping "active" in sync with the reporting.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticRatioGauge.java
udoprog: As the examples, having two `Counter`s would work, or one `Counter` and one `AtomicLong` (for total threads).

gabrielgerhardsson: Ok

gabrielgerhardsson:
Prepared fixes earlier today but am working on getting it tested on a test cluster before pushing.

gabrielgerhardsson:
@udoprog: Have worked on the parts you mentioned. What do you think?

Have tested the whole thing against Kafka with an extra piece of code that kills a ConsumerThread after a while, to simulate what would happen in the case of a message CRC error. The failure of a thread shows up in the metric as expected. Ran this test this morning, and then again now with the added fixes.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticRatioGauge.java
```diff
@@ -0,0 +1,68 @@
...
+    private final Counter numerator = new Counter();
+    private final Counter denominator = new Counter();
+
+    public SemanticRatioGauge(final long numerator, final long denominator) {
+        this.numerator.inc(numerator);
+        this.denominator.inc(denominator);
+    }
+
+    public void setNumerator(final long value) {
```
udoprog: Method doesn't appear to be used, or is it just included for completeness sake?

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticRatioGauge.java
```diff
@@ -0,0 +1,68 @@
...
+        numerator.inc();
+    }
+
+    public void decNumerator() {
+        numerator.dec();
+    }
+
+    public void setDenominator(final long value) {
+        denominator.dec(denominator.getCount());
+        denominator.inc(value);
```
udoprog: You can avoid these awkward blocks by making `numerator` and `denominator` an `AtomicLong`.

udoprog:
@gabrielgerhardsson very nice, just two more comments.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticRatioGauge.java
gabrielgerhardsson: Just included for completeness sake, since I wanted the class to be generically useful.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticRatioGauge.java
gabrielgerhardsson: Awesome, yes, it would feel much better. I chose Counter since it was mentioned earlier. I'll fix it :)

gabrielgerhardsson:
@udoprog: There you go :)

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/Units.java
```diff
@@ -35,4 +35,5 @@
     public static final String DROP = "drop";
     public static final String COUNT = "count";
     public static final String SAMPLE = "sample";
+    public static final String RATIO = "ratio";
```
udoprog: Just realized that the unit should be `%` for ratios.

udoprog: Yeah.

gabrielgerhardsson: Have fixed it now.

Issue #167, opened 2016-12-14
Amount of memory that a single aggregation is allowed to use should be determined by a quota


udoprog:
Currently it would be possible to construct a query for which the amount of memory required to serve it would be unsustainable.

Simple examples are the newly introduced topk/bottomk/abovek/belowk/pointsabove/pointsbelow, which all require that all data to be filtered has to be stored before they can be filtered.

Proposed solutions include:

* Implement a quota system.
* Implement a feature flag that requires the first aggregation step has to be a sampling aggregation, to reduce the amount of memory needed by downstream (filtering) aggregations. This is typically referred to as downsampling.

PR #184, opened 2017-01-04
Add aggregation quota

jo-ri:


jo-ri:
Draft implementation of #167 
@udoprog

aggregation/simple/src/test/java/com/spotify/heroic/aggregation/simple/FilterAggregationTest.java
```diff
@@ -26,31 +27,35 @@ public void setup() {
     public void testAboveKInstance() throws Exception {
         // TODO: support @JsonCreator
         // verifyClassBuilder(AboveKInstance.class).checkGetters(false).verify();
-        verifyRoundtrip("{\"type\":\"abovek\",\"k\":0.0}", new AboveKInstance(0),
+        verifyRoundtrip("{\"type\":\"abovek\",\"k\":0.0,\"inner\":{\"type\":\"EmptyInstance\"}}",
```
jo-ri: Fixed

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -389,15 +397,20 @@ private void accept(final Consumer<MetricBackend> op) {
...
         @Override
         public void resolved(final Pair<Series, FetchData> result) throws Exception {
             final FetchData f = result.getRight();
-
+            boolean failed = false;
             for (final MetricCollection g : f.getGroups()) {
-                g.updateAggregation(session, result.getLeft().getTags(),
-                    ImmutableSet.of(result.getLeft()));
+                failed = failed ||
+                    failOnLimits && aggregationLimit.isGreater(session::getRetainedMetricsCount);
```
udoprog: This approach would scan through all aggregations and sum up the retained count every time it's fed data.

As an alternative, we could use a system similar to the way that `QuotaWatcher` works. Each aggregation session would be initialized with an `AtomicLong` (wrapped behind an interface) and would be responsible for incrementing this every time it "stored" data. Querying it would only be a matter of reading its value here.

We might also want to consider tying these two systems together, since if the aggregation system "runs out of quota" there is no need to continue reading data from the backend any longer.

jo-ri: Implemented suggested alternative approach

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -155,12 +164,15 @@ public boolean isEmpty() {
             final AggregationInstance aggregation = request.getAggregation();
             final DateRange range = request.getRange();
 
-            final FetchQuotaWatcher delegateQuotaWatcher =
-                options.getDataLimit().orElse(dataLimit).asLong().<FetchQuotaWatcher>map(
-                    LimitedFetchQuotaWatcher::new).orElse(FetchQuotaWatcher.NO_QUOTA);
+            options.getDataLimit().orElse(dataLimit).asLong().orElse(Long.MAX_VALUE);
```
udoprog: no-op?

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -424,12 +440,23 @@ public FullQuery end(int resolved, int failed, int cancelled) throws Exception {
...
+
+            if (watcher.isRetainQuotaViolated()) {
+                limitsBuilder.add(ResultLimit.AGGREGATION);
+            }
+
+            if (watcher.isReadQuotaViolated()) {
+                limitsBuilder.add(ResultLimit.QUOTA);
+            }
+
+            if (watcher.isReadQuotaViolated() || failOnLimits && watcher.isRetainQuotaViolated()) {
```
udoprog: `failOnLimits` should not be checked for this quota

heroic-dist/src/test/java/com/spotify/heroic/AbstractClusterQueryIT.java
```diff
@@ -267,6 +267,26 @@ public void dataLimit() throws Exception {
     }
 
     @Test
+    public void aggregationLimit() throws Exception {
+        final QueryResult result = query("sum(10ms) by *", builder -> {
+            builder.options(Optional.of(QueryOptions.builder().failOnLimits(true).aggregationLimit(1L).build()));
```
udoprog: `failOnLimits` should not be needed here.

udoprog:
@jo-ri please squash your commits, then this should be ready for merging IMO.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -396,11 +407,15 @@ private void accept(final Consumer<MetricBackend> op) {
...
-                dataInMemoryReporter.reportDataNoLongerNeeded(g.size());
+            try {
+                for (final MetricCollection g : f.getGroups()) {
+                    g.updateAggregation(session, result.getLeft().getTags(),
+                        ImmutableSet.of(result.getLeft()));
+                    dataInMemoryReporter.reportDataNoLongerNeeded(g.size());
+                }
+            } catch (QuotaViolationException e) {
+                // Seems resolved() should not throw exception
+                log.debug("Aggregation quota violated", e);
```
udoprog: This is actually better if appended to `requestErrors`, then it would be part of any local stacktraces (as suppressed exceptions).

jo-ri: Wouldn't that result in an arbitrary number of requestErrors returned? Or do you mean add it to `errors`? Would that have any effect given that the collecting of errors is not reached when quota is violated?

jo-ri:
Squashed all commits

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
udoprog: Hm, at most as many errors as there are fetches under the current design? But for now I'm ok with this due to them not being propagated below regardless.

PR #185, opened 2017-01-04
Propagate request errors

jo-ri:


udoprog:
As a general note, it's useful if you prefix the commit message with the modules it touches. Or `[all]` if more than three.

PR #187, opened 2017-01-04
[core] Use the new recursion safety feature in tiny-async

gabrielgerhardsson:
This makes DelayedCollectCoordinator able to handle when there's many AsyncFuture's linked that are immediately resolved. Previous to this change, that scenario would result in a stack overflow.

This requires tiny-async 1.11.0, which we just switched to.


gabrielgerhardsson:
@udoprog: I can add some configuration or feature flag, if you'd like?

udoprog:
> @udoprog: I can add some configuration or feature flag, if you'd like?

No, let's go with this. We've tested the implementation of it fairly well already. If we find that it causes any issues we can hide it behind configuration and refine it.

PR #191, opened 2017-01-11
Add metadata-write shell task

jo-ri:


heroic-core/src/main/java/com/spotify/heroic/shell/task/MetadataWrite.java
```diff
@@ -0,0 +1,98 @@
...
+import javax.inject.Inject;
+import javax.inject.Named;
+
+import dagger.Component;
+import lombok.ToString;
+
+@TaskUsage("Write metadata")
+@TaskName("metadata-write")
+public class MetadataWrite implements ShellTask {
+
```
udoprog: Generally we only use blank lines between member and methods of a class, which should be in line with the Google code style.

PR #194, opened 2017-01-16
[RFC] Detailed query logging

parmus:
[Rendered](https://github.com/spotify/heroic/blob/rfc/detailed-query-logging/docs/rfcs/0000-detailed-query-logging.md)


gabrielgerhardsson:
Rebased on master

PR #196, opened 2017-01-18
Detailed Quota Limits

udoprog:
[Rendered](https://github.com/udoprog/heroic/blob/rfc/detailed-quota-limit/docs/rfcs/0000-detailed-quota-limits.md)


gabrielgerhardsson:
LGTM

PR #200, opened 2017-01-19
Introduce modules to allow for granular dependencies

### LINKED ISSUE #110: Build FakeModuleLoader that simplifies test bootstrapping for many modules
I want to turn [this](https://github.com/spotify/heroic/blob/master/aggregation/simple/src/test/java/com/spotify/heroic/aggregation/simple/FilterAggregationTest.java):

``` java
mapper = new ObjectMapper();
mapper.addMixIn(AggregationInstance.class, TypeNameMixin.class);
mapper.registerSubtypes(new NamedType(AboveKInstance.class, AboveK.NAME));
mapper.registerSubtypes(new NamedType(BelowKInstance.class, BelowK.NAME));
mapper.registerSubtypes(new NamedType(BottomKInstance.class, BottomK.NAME));
mapper.registerSubtypes(new NamedType(TopKInstance.class, TopK.NAME));
```

Into this:

``` java
final FakeModuleLoader loader = FakeModuleLoader.builder().module(Module.class).build();
mapper = loader.json();
```

[`Module`](https://github.com/spotify/heroic/blob/master/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/Module.java) already maps out the type names. The reason why it can't be used in the tests are because it requires the loading phase of Heroic to be configured.

In this, I propose to introduce `FakeModuleLoader` which sets up a fake loading environment and loads the specified module in order to provide a correctly configured `ObjectMapper`.

udoprog:
This PR changes the base structure of the project by moving loading and all its dependencies into a separate component.

By doing this, it will be easier for modules to depend on the loading phase of Heroic when writing tests, this will be the default when depending on `heroic-test`.

The new structure will allow for all modules to add a _testing_ dependency to `heroic-test`, which gives access to basic infrastructure to greatly simplify test fixtures.

This fixes #110


heroic-core/src/main/java/com/spotify/heroic/HeroicCore.java
```diff
@@ -613,7 +611,10 @@ private HeroicConfig config(LoadingComponent loading) throws Exception {
             builder = builder.merge(fragment);
         }
 
-        final ObjectMapper mapper = loading.configMapper();
+        final ObjectMapper mapper = loading.configMapper().copy();
+
+        // TODO: figure out where to put this
```
gabrielgerhardsson: Shall we do something about this now, or do it later?
If it's not critical, I vote for doing incremental changes and thus letting this be for now.

udoprog: :+1: for not dealing with it right now

udoprog:
@gabrielgerhardsson I pruned all the commits that introduce the change (they are now in the `use-fake-module-loader` branch). That should make it more clear what this does so others can have a look as well. I'll setup a separate PR for the other stuff once this is ready.

PR #208, opened 2017-01-29
[docs/rfcs] Aligned Timestamps

udoprog:
Strongly define how time buckets are aligned, specifically with an half-open interval `[start, end)`.

[Rendered](https://github.com/udoprog/heroic/blob/rfc/aligned-timestamp/docs/rfcs/0000-aligned-timestamp.md)


docs/rfcs/0000-aligned-timestamp.md
```diff
@@ -0,0 +1,68 @@
...
+
+Current fetches for `BigtableBackend` have one edge case for every period interval where data is invisible (see: https://github.com/udoprog/heroic/blob/fake-bigtable-backend/heroic-test-it/src/main/java/com/spotify/heroic/test/AbstractMetricBackendIT.java#L129).
+An interval definition should be picked which makes this easier to avoid.
+
+## Suggestion
+
+The current time buckets have values at timestamp `t` include the range `(t - extent, t]` in each
+bucket.
+
+Define a `DateRange` to be `[start, end)`. Buckets at timestamp `t`, should include data in the
```
jo-ri: Does this refer to a class name? If so, maybe better to keep it outside the RFC to avoid going into implementation.

udoprog: Changed it to be more explicit that `DateRange` refers to current implementation.

gabrielgerhardsson:
LGTM

PR #210, opened 2017-02-02
[metric] Use utf-8 in MetricsRowKeySerializerTest

zalenski:
@udoprog


metric/bigtable/src/test/java/com/spotify/heroic/metric/bigtable/MetricsRowKeySerializerTest.java
```diff
@@ -1,5 +1,6 @@
 package com.spotify.heroic.metric.bigtable;
 
+import com.google.common.base.Charsets;
```
udoprog: Nowadays there is `java.nio.charset.StandardCharsets`.

PR #214, opened 2017-02-07
[component, core] Add httpContext to all query log types

gabrielgerhardsson:
httpContext contains information about the http connection from where the query
originated from. It was previously only logged for the 'http-query-*' query
types. It's now added to all types so that the information is always available
for filtering in kibana.


gabrielgerhardsson:
@udoprog

gabrielgerhardsson:
@zfrank: I've merged this, but it would still be cool if you looked at the code as suggested.

PR #219, opened 2017-02-13
[heroic] Transactional Kafka consumer

gabrielgerhardsson:
Ensure safe Kafka consumption by disabling Kafka auto-commit and instead
explicitly commiting for data that we know has been completely written
to backend.

More specifically, we periodically pause consumption, wait for all
outstanding backend write requests to complete, then commit Kafka
offsets.

### DIFF
--- a/consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/ConsumerThread.java
+++ b/consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/ConsumerThread.java
@@ -24,23 +24,28 @@ package com.spotify.heroic.consumer.kafka;
 import com.spotify.heroic.consumer.ConsumerSchema;
 import com.spotify.heroic.consumer.ConsumerSchemaValidationException;
 import com.spotify.heroic.statistics.ConsumerReporter;
+import com.spotify.heroic.statistics.FutureReporter;
 import eu.toolchain.async.AsyncFramework;
 import eu.toolchain.async.AsyncFuture;
 import eu.toolchain.async.ResolvableFuture;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.CountDownLatch;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.atomic.LongAdder;
+import kafka.consumer.ConsumerTimeoutException;
 import kafka.consumer.KafkaStream;
 import kafka.message.MessageAndMetadata;
+import lombok.Setter;
 import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public final class ConsumerThread extends Thread {
-    private static final long INITIAL_SLEEP = 5;
-    private static final long MAX_SLEEP = 40;
+    private static final long RETRY_INITIAL_SLEEP = 5;
+    private static final long RETRY_MAX_SLEEP = 40;
 
     private final AsyncFramework async;
     private final String name;
@@ -50,6 +55,7 @@ public final class ConsumerThread extends Thread {
     private final AtomicInteger active;
     private final AtomicLong errors;
     private final LongAdder consumed;
+    private long threadLocalSequenceNumber = 0;
     // use a latch as a signal so that we can block on it instead of Thread#sleep (or similar) which
     // would be a pain to
     // interrupt.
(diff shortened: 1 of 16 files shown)


heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100.java
```diff
@@ -169,7 +170,12 @@ public void consume(final byte[] message) throws ConsumerSchemaException {
             final List<Point> points = ImmutableList.of(p);
 
             reporter.reportMessageDrift(System.currentTimeMillis() - p.getTimestamp());
-            ingestion.write(new Ingestion.Request(series, MetricCollection.points(points)));
+            AsyncFuture<Ingestion> ingestionFuture =
+                ingestion.write(new Ingestion.Request(series, MetricCollection.points(points)));
+
+            // Return Void future, to not leak unnecessary information from the backend but just
+            // allow monitoring of when the consumption is done.
+            return ingestionFuture.directTransform(future -> (Void) null);
```
udoprog: cast should not be necessary

consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/OffsetCommitThread.java
```diff
@@ -0,0 +1,182 @@
...
+    private void waitForOutstandingWrites() {
+        // At this point all ConsumerThread's are paused
+        log.info("All ConsumerThreads are paused (took " +
+            fullOperationWatch.elapsed(TimeUnit.MILLISECONDS) + " ms)");
+        Stopwatch waitForWritesWatch = Stopwatch.createStarted();
+
+        // Now, find all outstanding writes and wait for them to complete
+        List<AsyncFuture<Void>> outstandingWrites = new ArrayList<>();
+        for (ConsumerThread t : consumerThreads) {
+            AsyncFuture<Void> maybe = t.getLatestConsumptionFuture();
```
udoprog: It is important to note that updates to `latestConsumptionFuture` might not be visible to other threads.
In order to guarantee that writes to the `latestIngestionFuture` field are visible across threads we must establish a happens-before relationship. This can be done through synchronization, or with the use of `volatile` or something equivalent - like writing/reading an AtomicInteger.

For details, see: https://www.cs.umd.edu/~pugh/java/memoryModel/jsr-133-faq.html

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100.java
gabrielgerhardsson: 👍

consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/OffsetCommitThread.java
gabrielgerhardsson: Ah right, important indeed! Thank you.

gabrielgerhardsson:
Thanks for the review @udoprog!
I chose the current solution to incur minimal impact on performance. However, your suggestion would be much more straightforward indeed, so perhaps it's worth a performance hit. I'll try it - let's see what performance we get :)

consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/KafkaConsumerModule.java
```diff
@@ -61,20 +51,37 @@
...
+import lombok.Data;
+import lombok.NoArgsConstructor;
+import lombok.RequiredArgsConstructor;
+import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 @Data
 public class KafkaConsumerModule implements ConsumerModule {
     public static final int DEFAULT_THREADS_PER_TOPIC = 2;
+    private static final Boolean DEFAULT_TRANSACTIONAL = true;
```
udoprog: default should be `false` for now.

consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/ConsumerThread.java
```diff
@@ -195,9 +245,40 @@ private void consumeOneWithRetry(final byte[] body) throws InterruptedException
...
+
+            // FIXME: How does this work with batch writes?
+            threadLocalSequenceNumber++;
+            final Long sequenceNumber = new Long(threadLocalSequenceNumber);
+            if (enablePeriodicCommit) {
+                outstandingConsumptionRequests.add(sequenceNumber);
+            }
+
+            future.onFinished(() -> {
+                if (enablePeriodicCommit) {
```
udoprog: check if-statement outside of assigning future to avoid overhead if disabled.

consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/ConsumerThread.java
```diff
@@ -74,6 +95,13 @@ public ConsumerThread(
         this.active = active;
         this.errors = errors;
         this.consumed = consumed;
+        this.enablePeriodicCommit = enablePeriodicCommit;
+        this.outstandingConsumptionRequests = Collections.newSetFromMap(new ConcurrentHashMap());
```
udoprog: use of raw class, should be `new ConcurrentHashMap<Long, Boolean>()`

consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/ConsumerThread.java
```diff
@@ -195,9 +245,40 @@ private void consumeOneWithRetry(final byte[] body) throws InterruptedException
...
+            maybePrepareToCommitConsumerOffsets();
+
+            final FutureReporter.Context consumptionContext = reporter.reportConsumption();
+
+            // Actually consume
+            final AsyncFuture<Void> future = schema.consume(body);
+
+            // FIXME: How does this work with batch writes?
+            threadLocalSequenceNumber++;
+            final Long sequenceNumber = new Long(threadLocalSequenceNumber);
```
udoprog: this is not needed, see:
https://docs.oracle.com/javase/tutorial/java/data/autoboxing.html

consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/KafkaConsumerModule.java
gabrielgerhardsson: Fixed, thanks

consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/ConsumerThread.java
gabrielgerhardsson: Fixed

consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/ConsumerThread.java
gabrielgerhardsson: Fixed

consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/ConsumerThread.java
gabrielgerhardsson: Fixed

PR #222, opened 2017-02-20
[all] add credential=default to BigtableProfile

udoprog:


heroic-all/src/main/java/com/spotify/heroic/profile/BigtableProfile.java
```diff
@@ -67,7 +68,9 @@
                 module.credentials(new ComputeEngineCredentialsBuilder());
                 break;
             default:
-                throw new IllegalArgumentException("credentials: invalid value: " + credentials);
```
zalenski: keep the default IllegalArgumentException!

zalenski: done

zalenski:
(adding default credentials to BigtableAnalyticsProfile too in #252)

PR #223, opened 2017-02-24
[consumer] Rename a couple of things in ConsumerThread

gabrielgerhardsson:
No functionality change, only renaming.


consumer/kafka/src/main/java/com/spotify/heroic/consumer/kafka/ConsumerThread.java
```diff
@@ -129,58 +129,58 @@ private void threadIsStopping() {
         return async.resolved();
     }
 
-    public boolean isPaused() {
-        return this.paused.get() != null;
+    public boolean getShouldPause() {
```
udoprog: `is*` is convention for accessing booleans,  so `isPaused()` or `isPausing()` might be better names here..

gabrielgerhardsson: Thanks. That was a left over from experiments in another branch where I made some other changes to the pause logic. I've fixed it now.

PR #224, opened 2017-03-01
[component,core,metric/datastax] Fix API regression where .options couldn't be specified

gabrielgerhardsson:
…uldn't be specified


gabrielgerhardsson:
@udoprog

PR #226, opened 2017-03-03
[Metadata] Pick the correct indexes

parmus:
The first commit fixes a bug, which prevents Heroic from loading historical data, when querying with an absolute range in the past.

The two following commits just cleans up the API.


heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/index/RotatingIndexMapping.java
```diff
@@ -120,7 +120,7 @@ public String template() {
 
     @Override
     public String[] readIndices(DateRange range) throws NoIndexSelectedException {
-        return readIndices(range.end());
+        return readIndices(System.currentTimeMillis());
```
gabrielgerhardsson: I know some parts of how we do the indexing on a high level, but when I look at this code I realise that it's hard to understand just from the code. Throwing away the range and just passing in the current time seems illogical to any new developer looking at the code, so I think it would be great to have some explanation around that.

PR #229, opened 2017-03-10
[Aggregation] Make sure that aggregation cadence is not lost in distributed reduction.

parmus:
This patch is attempting to make sure that queries containing a delta aggregation still returns a meaningful cadence.

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/aggregation/DistributedAggregationCombiner.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/aggregation/DistributedAggregationCombiner.java
@@ -33,6 +33,20 @@ import java.util.List;
 public class DistributedAggregationCombiner implements AggregationCombiner {
     private final AggregationInstance reducer;
     private final DateRange range;
+    private final long cadence;
+
+    /**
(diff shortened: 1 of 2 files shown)


parmus:
@udoprog: Please review. We are not 100% sure we're fixing this in the right way.

udoprog:
@parmus to make this consistent it should probably be solved by propagating the current cadence similarly to how required tags are currently being propagated:
https://github.com/spotify/heroic/blob/master/heroic-component/src/main/java/com/spotify/heroic/aggregation/Chain.java

In effect add the 'last cadence' to the aggregation context and have the delta aggregation inherit that. Aggregations that change the cadence would do so appropriately by causing the context to be modified.

AggregationContext also appears to have bit-rotted a bit, `defaultSize` and `defaultExtent` should both be collapsed into a field called `lastCadence` or similarly. This is based of `defaultExtent` never being used which effectively makes `defaultSize` == `lastCadence`.

udoprog:
I forgot to mention that allowing the cadence to travel with the chain would permit us to _not_ having to specify the sampling in every step when sending the query to the API. Something UI developers and users would benefit from.

PR #230, opened 2017-03-11
[component] fix sampling in bucket aggregations

udoprog:
This modifies bucket aggregations so that they don't include the first bucket in the range since it
will never be provided data due to how it is queried. Data fetches are performed in the range
`(start - extent, start]`, but buckets are setup in the range `[start, end]` with size being used as
a distance between buckets.

Since the bucket selection algorithm needed to change because of this, it was simplified and test
cases added to clarify how it works.

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/aggregation/BucketAggregationInstance.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/aggregation/BucketAggregationInstance.java
@@ -28,27 +28,24 @@ import com.google.common.collect.Iterables;
 import com.spotify.heroic.common.DateRange;
 import com.spotify.heroic.common.Series;
 import com.spotify.heroic.common.Statistics;
-import com.spotify.heroic.metric.Payload;
 import com.spotify.heroic.metric.Event;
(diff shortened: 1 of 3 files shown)


heroic-component/src/main/java/com/spotify/heroic/aggregation/BucketAggregationInstance.java
```diff
@@ -267,4 +252,13 @@ public String toString() {
     private interface BucketConsumer<B extends Bucket, M extends Metric> {
         void apply(B bucket, M metric);
     }
+
+    /**
+     * A start, and an end bucket (exclusive) selected.
```
gabrielgerhardsson: Why exclusive here? 
I'm thinking that it would be more intuitive to simply specify the indices of the first and last (inclusive) buckets in the wanted bucket range - The bucket range that will include all samples for the wanted time range [start, end)

udoprog: To support the plain old for loop:
```java
for (int i = start; i < end; i++) {
    /* do something */
}
```

PR #244, opened 2017-03-23
[all] A shard with no discovered nodes is an error

gabrielgerhardsson:
This requires configuration - reusing the already existing 'topology'
configuration variable. Use it to specify which shards that are
expected to exist / be represented by nodes in the cluster. If Heroic
at some point has zero known nodes for any of these shards that will
be treated as a shard error and shown in query responses in the form
of entries in the 'errors' hash.
The list of expected shards is optional, but can be specified as so:

```
    cluster:
      tags:
        site: site2
        part: a
      discovery:
        type: ...
      topology:
        - site: site1
        - site: site2
          part: a
        - site: site2
          part: b
        - site: site3
```

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterDiscovery.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterDiscovery.java
@@ -22,7 +22,6 @@
 package com.spotify.heroic.cluster;
 
 import eu.toolchain.async.AsyncFuture;
-
 import java.net.URI;
 import java.util.List;
 
--- a/heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java
@@ -35,13 +35,6 @@ import eu.toolchain.async.AsyncFramework;
 import eu.toolchain.async.AsyncFuture;
 import eu.toolchain.async.LazyTransform;
 import eu.toolchain.async.Transform;
-import lombok.Data;
-import lombok.ToString;
-import lombok.extern.slf4j.Slf4j;
-import org.apache.commons.lang3.tuple.Pair;
-
-import javax.inject.Inject;
-import javax.inject.Named;
 import java.net.URI;
 import java.util.ArrayList;
 import java.util.Collection;
@@ -57,6 +50,12 @@ import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
+import javax.inject.Inject;
+import javax.inject.Named;
+import lombok.Data;
+import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
+import org.apache.commons.lang3.tuple.Pair;
 
 /**
  * Handles management of cluster state.
@@ -81,6 +80,7 @@ public class CoreClusterManager implements ClusterManager, LifeCycles {
     private final HeroicConfiguration options;
     private final LocalClusterNode local;
     private final HeroicContext context;
+    private final Set<Map<String, String>> expectedTopology;
 
     final AtomicReference<Set<URI>> staticNodes = new AtomicReference<>(new HashSet<>());
     final AtomicReference<NodeRegistry> registry = new AtomicReference<>();
@@ -93,7 +93,7 @@ public class CoreClusterManager implements ClusterManager, LifeCycles {
         AsyncFramework async, ClusterDiscovery discovery, NodeMetadata localMetadata,
         Map<String, RpcProtocol> protocols, Scheduler scheduler,
         @Named("useLocal") Boolean useLocal, HeroicConfiguration options, LocalClusterNode local,
-        HeroicContext context
+        HeroicContext context, @Named("topology") Set<Map<String, String>> expectedTopology
     ) {
         this.async = async;
         this.discovery = discovery;
@@ -104,6 +104,7 @@ public class CoreClusterManager implements ClusterManager, LifeCycles {
         this.options = options;
         this.local = local;
         this.context = context;
+        this.expectedTopology = expectedTopology;
     }
 
     @Override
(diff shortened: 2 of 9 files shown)


udoprog:
This seems like something that should live in the cluster section, since it does not depend on anything specific to the discovery implementation.

`expectedSites` should probably be `expectedShards`, since a shard is defined by _all_ of its tags:

```yaml
cluster:
  tags:
    site: site1
    part: a
  expectedShards:
    - site: site1
      part: a
    - site: site1
      part: b
```

This would still work very similarly to what has been done in this PR without being opinionated about the exact keys of the tags used.

gabrielgerhardsson:
@udoprog Thanks, I learnt something new here :) I didn't realise that several shards could contain the same site, with additional tags to differentiate them. Indeed, as you say, the way you structure the configuration the code wouldn't have to know which tags to look for. Very nice.

gabrielgerhardsson:
@udoprog: Me and @zfrank did the change you requested. So this PR is now ready for code review again.

heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterDiscovery.java
```diff
@@ -22,10 +22,14 @@
...
 import java.net.URI;
 import java.util.List;
+import java.util.Optional;
 
 public interface ClusterDiscovery {
     AsyncFuture<List<URI>> find();
+
+    default Optional<List<String>> expectedSites() {
+        return Optional.empty();
+    }
```
udoprog: Trailing change.

heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java
```diff
@@ -274,7 +276,33 @@ public void register(LifeCycleRegistry registry) {
...
+                .collect(Collectors.toList()));
+
+            final List<Map<String, String>> shardsWithNoNodes = new ArrayList<>();
+            for (final Map<String, String> expected : expectedShards.get()) {
+                if (!foundShards.contains(expected)) {
+                    shardsWithNoNodes.add(expected);
+                }
+            }
+
+            if (shardsWithNoNodes.size() > 0) {
```
udoprog: Dead statement? Loop will not loop if empty.

heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java
```diff
@@ -274,7 +276,33 @@ public void register(LifeCycleRegistry registry) {
...
+                for (final Map<String, String> shard : shardsWithNoNodes) {
+                    /* For every shard that didn't have a discovered node, add an empty entry in the
+                     * shard list. This ensures that suitable code later on will complain that there
+                     * were shards with no available nodes/groups. */
+                    shards.add(Pair.of(shard, ImmutableList.of()));
```
udoprog: Don't modify things returned from other APIs, they could easily be immutable. Instead, copy.

heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java
```diff
@@ -274,7 +276,33 @@ public void register(LifeCycleRegistry registry) {
...
+                .stream()
+                .map(mapListPair -> mapListPair.getKey())
+                .collect(Collectors.toList()));
+
+            final List<Map<String, String>> shardsWithNoNodes = new ArrayList<>();
+            for (final Map<String, String> expected : expectedShards.get()) {
+                if (!foundShards.contains(expected)) {
+                    shardsWithNoNodes.add(expected);
+                }
+            }
```
udoprog: Can be written like this using streams:
```java
final List<Map<String, String>> shardsWithNoNodes = expectedShards.get().stream().filter(e -> !foundShards.contains(e)).collect(Collectors.toList());
```

heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java
```diff
@@ -274,7 +276,33 @@ public void register(LifeCycleRegistry registry) {
...
+            registry.findFromAllShards(OptionalLimit.empty());
+
+        if (expectedShards.isPresent()) {
+            final Set<Map<String, String>> foundShards = new HashSet<>(shards
+                .stream()
+                .map(mapListPair -> mapListPair.getKey())
+                .collect(Collectors.toList()));
+
+            final List<Map<String, String>> shardsWithNoNodes = new ArrayList<>();
+            for (final Map<String, String> expected : expectedShards.get()) {
```
udoprog: With Optionals, prefer unwrapping early to a local variable to make it more obvious what it is.

As an alternative, this block can be rewritten as:

```java
return expectedShards.map(expected -> {
  /* do something with expected */
}).orElse(shards);
```

heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java
```diff
@@ -276,30 +275,32 @@ public void register(LifeCycleRegistry registry) {
...
+        if (shardsWithNoNodes.size() > 0) {
+            List<Pair<Map<String, String>, List<ClusterNode>>> updated = new ArrayList<>();
+            updated.addAll(shards);
+            for (final Map<String, String> shard : shardsWithNoNodes) {
+                /* For every shard that didn't have a discovered node, add an empty entry in the
+                 * shard list. This ensures that suitable code later on will complain that there
+                 * were shards with no available nodes/groups. */
+                updated.add(Pair.of(shard, ImmutableList.of()));
             }
+            shards = updated;
```
udoprog: branch can be inverted, and immediately return `shards` (which can then be final). In the other case, just return updated.

This also has the benefit of the branch being smaller.

```
if (shardsWithNoNodes.isEmpty()) {
    return shards;
}

/* build and return updated */
```

heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java
```diff
@@ -276,30 +275,32 @@ public void register(LifeCycleRegistry registry) {
...
-                     * were shards with no available nodes/groups. */
-                    shards.add(Pair.of(shard, ImmutableList.of()));
-                }
+        /* Topology (shards) is detected based on the metadata coming from the nodes. Any shards
+         * mentioned in 'topology' are required (shard error is raised if it isn't), additional
+         * shards may also exist tho. */
+        final Set<Map<String, String>> foundShards =
+            new HashSet<>(shards.stream().map(Pair::getKey).collect(Collectors.toList()));
+
+        log.info("expectedTopology: " + expectedTopology);
```
udoprog: rogue log statement?

heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java
gabrielgerhardsson: Indeed

heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java
gabrielgerhardsson: Argh :)

PR #249, opened 2017-03-31
[aggregation/simple,component,core] introduce BucketStrategy

udoprog:
This moves the specific implementation for how a bucket strategy is performed into the `BucketStrategy` interface.

This is intended to set us up so that we can eventually switch from the current `(end - extent, end]` (`BucketStrategy.END`) bucket interval, to `[start, start + extent)` (`BucketStrategy.START`).

To maintain backwards compatibility when the alternative implementation is introduced, a new feature (`com.spotify.heroic.end_bucket_strategy`) is introduced and enabled by default.

### DIFF
--- a/heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
@@ -29,10 +29,12 @@ import com.spotify.heroic.aggregation.AggregationInstance;
 import com.spotify.heroic.aggregation.AggregationOutput;
 import com.spotify.heroic.aggregation.AggregationResult;
 import com.spotify.heroic.aggregation.AggregationSession;
+import com.spotify.heroic.aggregation.BucketStrategy;
 import com.spotify.heroic.aggregation.RetainQuotaWatcher;
 import com.spotify.heroic.async.AsyncObservable;
 import com.spotify.heroic.common.DateRange;
 import com.spotify.heroic.common.Feature;
+import com.spotify.heroic.common.Features;
 import com.spotify.heroic.common.GroupSet;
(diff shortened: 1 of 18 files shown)


heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -163,17 +165,22 @@ public boolean isEmpty() {
...
             final QueryContext queryContext = request.getContext();
+            final Features features = request.getFeatures();
 
             queryLogger.logIncomingRequestAtNode(queryContext, request);
 
-            final boolean slicedFetch = request.getFeatures().hasFeature(Feature.SLICED_DATA_FETCH);
+            final boolean slicedFetch = features.hasFeature(Feature.SLICED_DATA_FETCH);
+
+            final BucketStrategy bucketStrategy =
+                options.getBucketStrategy().orElse(BucketStrategy.END);
```
gabrielgerhardsson: I believe you want to use features.hasFeature(Feature.END_BUCKET) here?

udoprog: There's no other bucket strategy right now. The new strategy and that check will be introduced in a separate PR.

gabrielgerhardsson: Yes I understand. It's just that Feature.END_BUCKET is unused as the code is right now. I would have expected it to be hooked up, just for completeness of the commit, even if the !hasFeature path would be just an exception.
No biggie tho of course :)

udoprog: It has been changed.

PR #251, opened 2017-04-04
[core] Explicitly close connection for failed nodes

gabrielgerhardsson:
When the cluster manager refreshes the cluster regularly, it will notice
if a node is down and remove it from the list of active nodes. This
works well. However, this commit ensures that the channel (containing
the actual tcp/ip connection) is also really closed. This fixes a
resource leak where sockets would not be closed.


zalenski:
LGTM

PR #252, opened 2017-04-04
[all] add credentials=default to BigtableAnalyticsProfile

zalenski:
Use it as default too.


zalenski:
@gabrielgerhardsson

zalenski:
(same as #222 but for BigtableAnalytics)

PR #261, opened 2017-04-24
Recover from RPC cancellation

gabrielgerhardsson:
Sometimes an RPC call to a node in the cluster may be cancelled. This
used to cause the whole query to be cancelled. This PR does two things:
* Adds metrics so that we know how often RPC calls gets errors and cancellations
* Now Heroic simply retries with the next node in the shard, when an RPC call was cancelled.

Possibly, the metrics should have their own reporter. However, I chose to put them in the same as the query metrics now, since it's quite related.

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterShard.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterShard.java
@@ -23,6 +23,7 @@ package com.spotify.heroic.cluster;
 
 import com.spotify.heroic.metric.QueryTrace;
 import com.spotify.heroic.metric.RuntimeNodeException;
+import com.spotify.heroic.statistics.QueryReporter;
 import eu.toolchain.async.AsyncFramework;
 import eu.toolchain.async.AsyncFuture;
 import eu.toolchain.async.RetryException;
@@ -47,6 +48,7 @@ public class ClusterShard {
 
     private final Map<String, String> shard;
     private final List<ClusterNode.Group> groups;
+    private final QueryReporter reporter;
 
     public <T> AsyncFuture<T> apply(
         Function<ClusterNode.Group, AsyncFuture<T>> function,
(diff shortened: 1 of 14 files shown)


heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterShard.java
```diff
@@ -77,10 +79,17 @@
             .retryUntilResolved(() -> {
                 final ClusterNode.Group next = it.next();
                 return function.apply(next).catchFailed(throwable -> {
+                    reporter.reportClusterNodeRPCError();
```
udoprog: Capitalization rules ignores acronyms, this method should be `reportClusterNodeRpcError()`. For the reasoning, consider the readability of HTTPXMLRPC vs HttpXmlRpc.

heroic-core/src/test/java/com/spotify/heroic/cluster/CoreClusterManagerTest.java
```diff
@@ -55,7 +56,7 @@ public void setUp() {
         final boolean useLocal = true;
 
         manager = spy(new CoreClusterManager(async, discovery, localMetadata, protocols, scheduler,
-            useLocal, options, local, context, ImmutableSet.of()));
+            useLocal, options, local, context, ImmutableSet.of(), NoopQueryReporter.get()));
```
udoprog: Mock `QueryReporter`, and possibly add some verifications to existing tests (or new tests)?

heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterShard.java
gabrielgerhardsson: Thanks for the reminder 😊  Fixed

heroic-core/src/test/java/com/spotify/heroic/cluster/CoreClusterManagerTest.java
gabrielgerhardsson: Fixed

gabrielgerhardsson:
@udoprog Thanks for the review, I've adressed the things you mentioned.

PR #263, opened 2017-04-26
Handle RPC cancellations

gabrielgerhardsson:
RPC calls that get cancelled unexpectedly should be handled. These commits make them be handled in a similar way as errors. Thus a cancelled 'metadata' RPC call in the cluster refresh will cause the node to be removed from the list of active nodes. Similarly for a cancelled initial 'metadata' RPC call directly after connect.
This also makes the grpc layer close the channel if the initial metadata RPC call is cancelled, just like it already does for errors. Potentially, this would otherwise be a rare tcp/ip socket leak.

### DIFF
--- a/heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java
@@ -46,6 +46,7 @@ import java.util.List;
 import java.util.Map;
 import java.util.Optional;
 import java.util.Set;
+import java.util.concurrent.CancellationException;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import java.util.concurrent.atomic.AtomicReference;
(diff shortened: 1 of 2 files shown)


gabrielgerhardsson:
@udoprog Thank you for your detailed thoughts!
"This is an example where cancellations should never happen" - I suspected that also, and @zalenski mention the same today. However, it *seems* like the grpc layer is actually cancelling RPC calls when there is connection issues. I don't have hard proof, but when looking at a large amount of logs this seems to be consistent: The https://github.com/spotify/heroic/commit/3aaf84eabcbc63c803e77871addfafbcddf47f55 commit links the "Request cancelled" error message to a specific shard. Adding some extra logging in that area made it possible to see that whenever a request was cancelled, there were also a connection problem soon before *or* after this, for that specific shard.

So, blindly adding handling of cancellation everywhere like this PR seeks to do is perhaps the wrong way forward..? Not a big deal tho, these would likely all be rare occasions. I prepared the PR as a defensive measure to cover all bases.

udoprog:
> So, blindly adding handling of cancellation everywhere like this PR seeks to do is perhaps the wrong way forward..?

No, this is still the right thing to. As mentioned, we need a when-not-completed cleanup method for most of these cases that includes both cancellations and errors. It just shouldn't have this much of an effect, so we also should figure out why the gRPC interaction has (many?) cancellations.

Please merge!

gabrielgerhardsson:
Ok thanks for the clarification 👍

PR #267, opened 2017-05-05
ES 5.3.2 migration

gabrielgerhardsson:
Integration tests and unit tests pass, and I added further tests for ScrollTransformStream.
This is a large change tho so perhaps I've missed something.

Especially curious why MetadataBackendV1 did a count for findTagKeys and findKeys. That specific way of using the API has been removed, so I'd need to understand what the intention was to mimic with ES 5.x. The integration tests pass, but perhaps we're retrieving unnecessary data now?
Or, we could just deprecate V1..


gabrielgerhardsson:
@udoprog: Would appreciate a code review on this.

mykolasmith:
How will this work re: backwards compatibility? Do you specify the ES version heroic.yml? How about migrating existing indices from 1.7?

udoprog:
@mykolasmith unfortunately it won't be backwards compatible due to https://github.com/elastic/elasticsearch/issues/13273

I will try to maintain a branch for ES 1.x, but the long term goal should be to migrate.

A rewrite to use the [rest client][rest-client] is not straight forward due to a lack of data models. It is very bare bones and didn't even exist when this project was started. But that could be a mid term solution as well as long as it doesn't affect performance negatively.

[rest-client]: https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/index.html

Another approach I've been experimenting with is to repackage the clients, but some dependencies make that awkward, and we would end up having to duplicated a bit of infrastructure (primarily [`elasticsearch-utils`][utils]) to make it work.

[utils]: https://github.com/spotify/heroic/tree/master/heroic-elasticsearch-utils

#### Migration Steps

As a start of possible documentation, these are the steps you could do to migrate:

This assumes that:
* You have a HTTP proxy in front of Heroic APIs.
* You are using Kafka to consume metrics.

1. Spin up another set of consumers that writes metadata into ES 5.
2. Spin up another set of APIs that uses the new code to talk with ES 5, but same metric backend, and put behind the proxy.
3. Validate the new setup and switch the proxy to talk with the new nodes.
4. Decommission the old infrastructure.

The [debian packaging][debian] is built in a way that you can easily generate a new non-conflicting package that can be installed on the same node, then it's just a matter of configuring the nodes.

[debian]: https://github.com/spotify/heroic#building-a-debian-package

gabrielgerhardsson:
@udoprog Updated with v1 schema removed (just skeleton remains).

mykolasmith:
FYI we've upgraded our QA cluster to 5.3.3 (accidentally went a minor version higher) and everything seems to be working swimmingly. Thanks for this much needed update! We're so happy to be on ES 5.x now, as index backup w/ Google Cloud Storage is now supported natively, meaning I'll rest easier at night :)

gabrielgerhardsson:
Happy to hear that @mykolasmith! :)

PR #268, opened 2017-05-05
Add tests to check for core cluster manager cancellations

udoprog:
This should not be merged until the problem has been resolved.

### DIFF
--- a/rpc/grpc/src/main/java/com/spotify/heroic/rpc/grpc/GrpcRpcProtocolServer.java
+++ b/rpc/grpc/src/main/java/com/spotify/heroic/rpc/grpc/GrpcRpcProtocolServer.java
@@ -60,6 +60,8 @@ import lombok.extern.slf4j.Slf4j;
 
 @Slf4j
 public class GrpcRpcProtocolServer implements LifeCycles {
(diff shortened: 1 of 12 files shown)


rpc/grpc/src/main/java/com/spotify/heroic/rpc/grpc/GrpcRpcProtocolServer.java
```diff
@@ -103,46 +102,61 @@ public GrpcRpcProtocolServer(
...
 
-        container.register(GrpcRpcProtocol.METADATA, empty -> async.resolved(localMetadata));
+        container.register(GrpcRpcProtocol.METADATA, empty -> {
+            return async.resolved(metadataProvider.getMetadata());
+        });
 
         container.register(GrpcRpcProtocol.METRICS_FULL_QUERY,
-            g -> g.apply(metrics, MetricBackendGroup::query));
+            g -> g.apply(metrics, MetricBackendGroup::query)
+        );
```
gabrielgerhardsson: This whitespace change and the ones below are cancelled in the next commit. Would be cleaner if the change is never introduced.

gabrielgerhardsson:
Cool stuff! :) We should have more like this 👌

udoprog:
Closed in favor of #269

PR #269, opened 2017-05-10
Mitigate RPC cancellations

gabrielgerhardsson:


gabrielgerhardsson:
@udoprog Merged in your changes in #268, together with one extra commit to make that work with the other changes in this PR.

heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterShard.java
```diff
@@ -47,16 +50,16 @@
     private final AsyncFramework async;
 
     private final Map<String, String> shard;
-    private final List<ClusterNode.Group> groups;
     private final QueryReporter reporter;
+    private final ClusterManager cluster;
+
+    private final List<UUID> nodesTried = new ArrayList<>();
```
udoprog: This should be inside `apply(..)`, otherwise its state would be shared across requests over one specific ClusterShard.

heroic-dist/src/test/java/com/spotify/heroic/cluster/CoreClusterManagerIT.java
```diff
@@ -0,0 +1,216 @@
...
+                        for (int p = 0; p < 10; p++) {
+                            final List<UUID> excludeIds = Lists.newArrayList();
+                            while (true) {
+                                // Send ping to all nodes in the shard
+                                Optional<ClusterManager.NodeResult<AsyncFuture<Void>>> ret =
+                                    clusterManager.withNodeInShardButNotWithId(shard.getShard(),
+                                        excludeIds, ClusterNode.Group::ping);
+                                if (!ret.isPresent()) {
+                                    // No more nodes available in shard
+                                    break;
```
udoprog: throw exception?

heroic-dist/src/test/java/com/spotify/heroic/cluster/CoreClusterManagerIT.java
```diff
@@ -0,0 +1,216 @@
...
+        while (!t.shutdown.get()) {
+            final List<Callable<AsyncFuture<Void>>> operations = new ArrayList<>();
+
+            for (int i = 0; i < requestsPerIteration; i++) {
+                operations.add(() -> {
+                    final List<AsyncFuture<Void>> pongs = new ArrayList<>();
+
+                    for (final ClusterShard shard : clusterManager.useDefaultGroup()) {
+                        for (int p = 0; p < 10; p++) {
+                            final List<UUID> excludeIds = Lists.newArrayList();
```
udoprog: The below is a bit messy, please put it in a helper method.

heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterManager.java
```diff
@@ -75,4 +109,14 @@
...
     Set<RpcProtocol> protocols();
+
+    /**
+     * The result of applying a function to a random node in a shard
+     */
+    @Data
+    class NodeResult<T> {
+        private final T returnValue;
+        private final UUID nodeId;
+        private final String nodeString;
```
udoprog: I like wrapping the result to provide more information, but maybe we should expose the `ClusterNode` directly here?

heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterManager.java
```diff
@@ -63,9 +65,41 @@
...
+     *
+     * @param shard the shard
+     * @param excludeIds a list of node Ids that will not be considered when choosing a node
+     * @param fn the function to apply on the node
+     * @param <T> type of the return value
+     * @return Return value of the applied function, Id of the node, string representation of node
+     */
+    <T> Optional<NodeResult<T>> withNodeInShardButNotWithId(
+        Map<String, String> shard, List<UUID> excludeIds, Function<ClusterNode.Group, T> fn
+    );
```
udoprog: Unless we have a specific optimization for this we can take the exclusion criteria as a `Predicate<ClusterNode> exclude` to make it less specific?

The call would (according to my napkin coding) then be:
```java
withNodeInShardButNotWithId(shard, excludeIds::contains, fn)
```

heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterShard.java
gabrielgerhardsson: Thanks, fixed!

heroic-dist/src/test/java/com/spotify/heroic/cluster/CoreClusterManagerIT.java
gabrielgerhardsson: Fixed

heroic-dist/src/test/java/com/spotify/heroic/cluster/CoreClusterManagerIT.java
gabrielgerhardsson: Now that it's in a helper I'm returning there instead.

Or did you object to the use of Optional<> as return value to indicate if there were any more nodes?

heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterManager.java
gabrielgerhardsson: Good point, fixed

heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterManager.java
gabrielgerhardsson: Fixed

gabrielgerhardsson:
@udoprog Thank you for the review. I've pushed an update now, with the following additions & changes:
* Reverted two of the three cancellation related commits that were done previously. These two should not be there, but the third one is ok in my opinion (as a safety measure against unexpected cancellations during cluster refresh).
* Code review changes in the new commit at the end. The commits in the middle have unchanged contents except for rebasing on top of reverts.

udoprog:
@gabrielgerhardsson Thanks for the changes.

Should we hold of on the revert until we've verified that the fix works. I also don't mind having that code in there since it's a defensive approach in general.

gabrielgerhardsson:
@udoprog Ok, holding the reverts 👍 
Removed the reverts and pushed again.

udoprog:
@gabrielgerhardsson Some checkstyle issues.

udoprog:
@gabrielgerhardsson Last part is just getting rid of the last commit, titled "Code review fixes. Squash before merge.".

gabrielgerhardsson:
@udoprog Thank you. Now squashed.

gabrielgerhardsson:
@udoprog thumbs up? :)

PR #270, opened 2017-05-12
Fix 'Heroic throws NullPointerException if it cannot find ElasticSearch index'

nezdolik:


nezdolik:
will push changes to remote branch and create new request

PR #274, opened 2017-05-18
Add metrics for metric backend reads

gabrielgerhardsson:
* query-metrics-samples-read - Per-query histogram of number of samples
  read from metric backend on data-node
* query-metrics-rows-accessed - Per-query histogram of number of rows in
  metric backend that was accessed on data-node. Can give an additional
  indication of the load that the query put on the backend.


gabrielgerhardsson:
Still a bit unsure if rows-accessed is calculated correctly. Please advice.

PR #275, opened 2017-05-18
[component,core,tools] Add rawSampleSize in query log

gabrielgerhardsson:
Expose in query response and query log a measurement of how many data points that were read from the metric backend. The measurement is done before aggregations etc, so gives an indication of how much load the query put on the metric backend.

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/metric/QueryMetricsResponse.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/metric/QueryMetricsResponse.java
@@ -32,10 +32,10 @@ import java.util.HashMap;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 import java.util.SortedSet;
 import java.util.UUID;
-
 import lombok.Data;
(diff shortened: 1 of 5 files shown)


heroic-component/src/main/java/com/spotify/heroic/metric/QueryMetricsResponse.java
```diff
@@ -84,6 +88,8 @@ public void serialize(
             g.writeFieldName("result");
             serializeResult(g, common, result);
 
+            g.writeObjectField("rawSampleSize", response.getPreAggregationSampleSize());
```
udoprog: Field should be named the same as the resulting JSON object to reduce confusion (if possible).

gabrielgerhardsson: True, it was a bit confusing. I was trying to avoid the long preAggregationSampleSize, but now I've pushed an update that uses that name.

udoprog: Ah, we could probably just use the short name. But this gives you the :+1:.

PR #282, opened 2017-05-22
Revert "[bigtable] Bump client version to 0.9.6.2"

gabrielgerhardsson:
Reverts spotify/heroic#281.

### DIFF
--- a/repackaged/bigtable/pom.xml
+++ b/repackaged/bigtable/pom.xml
@@ -10,7 +10,7 @@
   <groupId>com.spotify.heroic.repackaged</groupId>
   <artifactId>bigtable-client-core</artifactId>
   <packaging>jar</packaging>
-  <version>0.9.6.2</version>
+  <version>0.9.4</version>
 
   <name>Heroic: Re-Packaging of Bigtable Client Utilities</name>
 


udoprog:
@gabrielgerhardsson what's going on?

gabrielgerhardsson:
@udoprog The previous PR was premature, it turns out. I missed changing in a couple of other places, and when I do now it's not immediately solved. 

I could also just finish the work and merge as another fixup-PR. However, reverting and merging when it's really done seems better. 
What do you think @udoprog?

udoprog:
@gabrielgerhardsson What are the effects? But in general: fix it if quick, otherwise revert :).

gabrielgerhardsson:
@udoprog Building on my machine works fine, but automated build breaks. I'm still investigating. This is why I didn't catch it with a simple reinstall-repackaged && rebuild on my machine.

gabrielgerhardsson:
@udoprog Alright, I'll keep on looking at this and we'll see :)

gabrielgerhardsson:
@udoprog I believe this will require an upgrade of grpc, and a code change due to API change in the bigtable client. So I believe reverting and putting the change out there as a proper PR would be best.

PR #283, opened 2017-05-22
Bigtable / grpc / netty, version bumps

gabrielgerhardsson:
Upgrade bigtable client to 0.9.6.2, which requires updated grpc, which requires updated netty.

### DIFF
--- a/metric/bigtable/pom.xml
+++ b/metric/bigtable/pom.xml
@@ -22,7 +22,7 @@
   </description>
 
   <properties>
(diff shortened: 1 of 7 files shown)


gabrielgerhardsson:
rebuild

gabrielgerhardsson:
Note that this PR also has one possibly important code change: Bigtable will now use its own executor. The API to give it our own executor has been removed (it was already deprecated in the previous release - now it's gone).

PR #284, opened 2017-05-23
Add metrics for measuring query load

gabrielgerhardsson:
query-metrics-max-live-samples - Maximum amount of data points in memory
at any time for a query.

query-sample-latency - Effectively the latency of 1 sample read. This
makes for a latency measurement that is comparable across queries. This
can be used to show if any query loaded the system more than what would
have been expected from just the number of samples.

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/statistics/QueryReporter.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/statistics/QueryReporter.java
@@ -36,6 +36,9 @@ public interface QueryReporter {
      */
     void reportSmallQueryLatency(long duration);
 
+    void reportLatencyVsSize(long durationNs, long preAggregationSampleSize);
+
     void reportClusterNodeRpcError();
+
     void reportClusterNodeRpcCancellation();
 }
--- a/heroic-core/src/main/java/com/spotify/heroic/CoreQueryManager.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/CoreQueryManager.java
@@ -297,8 +297,10 @@ public class CoreQueryManager implements QueryManager {
             if (sampleSize > smallQueryThreshold) {
                 return;
             }
-            final long duration = fullQueryWatch.elapsed(TimeUnit.MILLISECONDS);
-            reporter.reportSmallQueryLatency(duration);
+            final long durationNs = fullQueryWatch.elapsed(TimeUnit.NANOSECONDS);
+            reporter.reportLatencyVsSize(durationNs, result.getPreAggregationSampleSize());
+            final long durationMs = TimeUnit.NANOSECONDS.toMillis(durationNs);
+            reporter.reportSmallQueryLatency(durationMs);
         }
 
         @Override
--- a/heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
@@ -21,6 +21,7 @@
 
 package com.spotify.heroic.metric;
 
+import com.google.common.collect.ConcurrentHashMultiset;
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.Iterables;
@@ -37,6 +38,7 @@ import com.spotify.heroic.common.Feature;
 import com.spotify.heroic.common.Features;
 import com.spotify.heroic.common.GroupSet;
 import com.spotify.heroic.common.Groups;
+import com.spotify.heroic.common.Histogram;
 import com.spotify.heroic.common.OptionalLimit;
 import com.spotify.heroic.common.QuotaViolationException;
 import com.spotify.heroic.common.SelectedGroup;
@@ -54,7 +56,6 @@ import eu.toolchain.async.AsyncFramework;
 import eu.toolchain.async.AsyncFuture;
 import eu.toolchain.async.LazyTransform;
 import eu.toolchain.async.StreamCollector;
-
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Optional;
@@ -65,7 +66,6 @@ import java.util.function.Consumer;
 import java.util.function.Function;
 import javax.inject.Inject;
 import javax.inject.Named;
-
 import lombok.RequiredArgsConstructor;
 import lombok.ToString;
 import lombok.extern.slf4j.Slf4j;
@@ -204,13 +204,12 @@ public class LocalMetricManager implements MetricManager {
 
                 if (result.isLimited()) {
                     if (failOnLimits) {
-                        final List<RequestError> errors = ImmutableList.of(QueryError.fromMessage(
+                        final RequestError error = QueryError.fromMessage(
                             "The number of series requested is more than the allowed limit of " +
-                                seriesLimit));
+                                seriesLimit);
 
-                        return async.resolved(
-                            new FullQuery(namedWatch.end(), errors, ImmutableList.of(),
-                                Statistics.empty(), ResultLimits.of(ResultLimit.SERIES)));
+                        return async.resolved(FullQuery.limitsError(namedWatch.end(), error,
+                            ResultLimits.of(ResultLimit.SERIES)));
                     }
 
                     limits = ResultLimits.of(ResultLimit.SERIES);
@@ -227,11 +226,11 @@ public class LocalMetricManager implements MetricManager {
                 try {
                     session = aggregation.session(range, quotaWatcher, bucketStrategy);
                 } catch (QuotaViolationException e) {
-                    return async.resolved(new FullQuery(namedWatch.end(), ImmutableList.of(
+                    return async.resolved(FullQuery.limitsError(namedWatch.end(),
                         QueryError.fromMessage(String.format(
                             "aggregation needs to retain more data then what is allowed: %d",
-                            aggregationLimit.asLong().get()))), ImmutableList.of(),
-                        Statistics.empty(), ResultLimits.of(ResultLimit.AGGREGATION)));
+                            aggregationLimit.asLong().get())),
+                        ResultLimits.of(ResultLimit.AGGREGATION)));
                 }
 
                 /* setup collector */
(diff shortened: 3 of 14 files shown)


gabrielgerhardsson:
Hold off on this one a bit. I'm adding a third metric.

heroic-core/src/main/java/com/spotify/heroic/CoreQueryManager.java
```diff
@@ -297,8 +297,11 @@ private void reportCompletedQuery(
             if (sampleSize > smallQueryThreshold) {
                 return;
             }
-            final long duration = fullQueryWatch.elapsed(TimeUnit.MILLISECONDS);
-            reporter.reportSmallQueryLatency(duration);
+            final long durationNs = fullQueryWatch.elapsed(TimeUnit.NANOSECONDS);
+            reporter.reportLatencyVsSize(durationNs, result.getPreAggregationSampleSize(),
+                result.getRange());
+            final long durationMs = TimeUnit.NANOSECONDS.toMillis(durationNs);
```
zalenski: `durationNs`?

heroic-component/src/main/java/com/spotify/heroic/statistics/QueryReporter.java
```diff
@@ -36,6 +38,8 @@
      */
     void reportSmallQueryLatency(long duration);
 
+    void reportLatencyVsSize(long duration, long preAggregationSampleSize, DateRange range);
```
zalenski: maybe `durationNs` here too, or document that it's nanoseconds somewhere?

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticMetricBackendReporter.java
```diff
@@ -147,16 +175,35 @@ public void reportDataNoLongerNeeded(final long n) {
...
+                queryMaxLiveSamples.update(maxDataInMemory.get());
+
+                long duration = queryWatch.elapsed(TimeUnit.NANOSECONDS);
+                if (duration != 0) {
+                    queryReadRate.update((1000_000_000 * dataReadLocal) / duration);
+                }
+
+                for (ReadRangeStatistics stats : seriesDensity.values()) {
+                    long rangeMs = stats.getRange().diff();
+                    queryRowDensity.update((3600_000L * stats.getValue().get()) / rangeMs);
```
zalenski: the unit for `queryRowDensity` is "count". where does the `3600_000L` constant come from? if it is _x_ per hour, would _x_ per second be more natural?

udoprog: Guessing here, but but a per-second value would probably round to zero in most cases. Most of our applications emits data around every 30 seconds IIRC.

Flipping the units might be prudent in this case (ms / sample). The highest theoretical density we support is 1ms, and we currently don't support non-long histograms.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticMetricBackendReporter.java
```diff
@@ -118,19 +131,34 @@ public MetricBackend decorate(final MetricBackend backend) {
...
         return new DataInMemoryReporter() {
-            private final AtomicLong dataRead = new AtomicLong(0);
-            private final AtomicLong dataInMemory = new AtomicLong(0);
-            private final AtomicLong rowsAccessed = new AtomicLong(0);
+            private final LongAccumulator dataRead = new LongAccumulator((x, y) -> x + y, 0L);
+            private final AtomicLong dataInMemory = new AtomicLong(0L);
+            private final LongAccumulator rowsAccessed = new LongAccumulator((x, y) -> x + y, 0L);
+            private final LongAccumulator maxDataInMemory = new LongAccumulator(Long::max, 0L);
+            private final Stopwatch queryWatch = Stopwatch.createStarted();
+            private final Map<Integer, ReadRangeStatistics> seriesDensity =
```
udoprog: This data structure really needs documentation.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticMetricBackendReporter.java
```diff
@@ -118,19 +131,34 @@ public MetricBackend decorate(final MetricBackend backend) {
...
             public void reportRowsAccessed(final long n) {
-                rowsAccessed.addAndGet(n);
+                rowsAccessed.accumulate(n);
+            }
+
+            @Override
+            public void reportSliceRead(
+                final Series series, final DateRange range, final long n
+            ) {
+                int key = series.hashCode() + range.hashCode();
```
udoprog: Hm, not sure if collisions are acceptable (10k-20k gives 1-5% for me). They would skew up the measurements. If you're concerned about key size, you can use the [`Series#getHashCode()`](https://github.com/spotify/heroic/blob/master/heroic-component/src/main/java/com/spotify/heroic/common/Series.java#L82) which is a 128-bit hash and has a low probability of collisions.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticMetricBackendReporter.java
gabrielgerhardsson: Thanks for your comments, I've already implemented your suggestion @udoprog - but haven't pushed that update yet since I'm also working on exposing the data in query log.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticMetricBackendReporter.java
gabrielgerhardsson: 👍

gabrielgerhardsson: Thanks.
I've been restructuring this part of the code. It's been a few days now since I touched the code so I don't have it fully in my head, but I believe my last experiment could manage without this data structure completely.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticMetricBackendReporter.java
gabrielgerhardsson: Thanks for the advice!
I've been restructuring this part of the code. It's been a few days now since I touched the code so I don't have it fully in my head, but I believe my last experiment could manage without this data structure completely. So no more risk for collisions in that case. If it turns out that this is still needed, I'll go with your proposal for sure.

heroic-component/src/main/java/com/spotify/heroic/statistics/QueryReporter.java
gabrielgerhardsson: 👍  Thanks for your comments.

gabrielgerhardsson:
@udoprog, @zalenski: This is now ready for code review, if you'd like. Thanks!

metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableBackend.java
```diff
@@ -472,8 +472,12 @@ public Statistics getStatistics() {
             fetches.add(readRows.directTransform(result -> {
                 for (final FlatRow row : result) {
                     watcher.readData(row.getCells().size());
-                    List<Metric> metrics = Lists.transform(row.getCells(), transform);
+                    final List<Metric> metrics = Lists.transform(row.getCells(), transform);
                     metricsConsumer.accept(MetricCollection.build(type, metrics));
+
+                    if (metrics.size() > 1) {
+                        watcher.reportRowDensity(getDensity(metrics));
```
udoprog: Two comments;

* It sucks that `metrics.size() == 0` is not well defined. That being said, I don't think we should just ignore it. It could be misrepresenting the data.
* Shouldn't this be done in `LocalMetricManager`? There's not specific dependency to BigtableBackend AFAICT.

gabrielgerhardsson: 1. Good point, I believe the code is there because of earlier reasons while I was experimenting - reasons that aren't valid any more. I'll move it to ResultCollector in LocalMetricManager.
1. metrics.size() == 0: So, throw an error and fail the whole read operation or add an entry to requestErrors?
1. metrics.size() == 1: How about this case? Can't calculate the density (distance between metrics) when there's only one metric. Is this a problem?

udoprog: Yeah, `1` as well. The desired goal would be to cause values to 'shift down' appropriately, so replacing them with something like `LONG.MAX_VALUE` could be a way.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -569,10 +577,11 @@ public FullQuery end(int resolved, int failed, int cancelled) throws Exception {
     }
 
     @RequiredArgsConstructor
-    private static class QuotaWatcher implements FetchQuotaWatcher, RetainQuotaWatcher {
+    private static class ReadWatcher implements FetchQuotaWatcher, RetainQuotaWatcher {
         private final long dataLimit;
         private final long retainLimit;
         private final DataInMemoryReporter dataInMemoryReporter;
+        private final ConcurrentHashMultiset<Long> rowDensityData = ConcurrentHashMultiset.create();
```
udoprog: Hate to be a stickler, but this should not live in the watcher anymore since it's not being reported by the backend :). A better place would be in the `ResultCollector`.

gabrielgerhardsson: :) You're very right. Fixed it now.

heroic-component/src/main/java/com/spotify/heroic/metric/FetchQuotaWatcher.java
```diff
@@ -31,6 +31,12 @@
     void readData(long n);
 
     /**
+     * Indicates the distance between samples in a row that was read from the metric backend
+     * @param msBetweenSamples
+     */
+    void reportRowDensity(long msBetweenSamples);
```
udoprog: This is no longer used?

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -566,10 +575,22 @@ public FullQuery end(int resolved, int failed, int cancelled) throws Exception {
...
+            for (final Long value : rowDensityData.elementSet()) {
+                builder.add(value);
+            }
+            return builder.build();
+        }
     }
 
     @RequiredArgsConstructor
-    private static class QuotaWatcher implements FetchQuotaWatcher, RetainQuotaWatcher {
+    private static class ReadWatcher implements FetchQuotaWatcher, RetainQuotaWatcher {
```
udoprog: This rename is unnecessary, and a bit misleading.

heroic-component/src/main/java/com/spotify/heroic/metric/FetchQuotaWatcher.java
gabrielgerhardsson: Very true, moved to ResultCollector now  👍

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
gabrielgerhardsson: You're right :) The benefit of being able to move other things to ResultCollector. I've changed it back to QuotaWatcher now.

PR #285, opened 2017-05-24
[heroic-core] enable gzip

mykolasmith:
Enabling gzip for responses from POST endpoints that serve JSON (e.g. `/query/metrics` and `/query/batch`) has a significant benefit when requesting large amounts of data. I chose 860 bytes for the minimum gzip size since that is the default that Akamai uses for their gzip threshold.


udoprog:
Thank you!

PR #286, opened 2017-05-30
[LocalMetricManager] Refactor transform into separate class

parmus:


heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -179,17 +179,24 @@ public boolean isEmpty() {
...
+                options.getDataLimit()
+                       .orElse(dataLimit)
+                       .asLong()
+                       .orElse(Long.MAX_VALUE),
+                options.getAggregationLimit()
+                       .orElse(aggregationLimit)
+                       .asLong()
+                       .orElse(Long.MAX_VALUE),
+                dataInMemoryReporter
+            );
```
gabrielgerhardsson: I agree that this looks better. However, the auto-formatting rules doesn't agree and will change this back the next time someone auto formats. The same is true for most of the other formatting changes in this file, except for the variable name changes (which I agree with).
So please either change the auto formatting rules somehow, or auto format before committing.

heroic-component/src/main/java/com/spotify/heroic/metric/FetchData.java
```diff
@@ -92,12 +92,10 @@ public static FetchData of(
...
-                    ImmutableList.Builder<Metric> data = fetchGroups.get(g.getType());
-
-                    if (data == null) {
-                        data = new ImmutableList.Builder<>();
-                        fetchGroups.put(g.getType(), data);
-                    }
+                    ImmutableList.Builder<Metric> data = fetchGroups.computeIfAbsent(
+                        g.getType(),
+                        k -> new ImmutableList.Builder<>()
+                    );
```
gabrielgerhardsson: 👍

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
parmus: This is a perfect example of what we're discussing in the channel right now. We choose to obey a tool and some rules rather than aim for readability.

parmus:
Fuck it

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -233,62 +241,94 @@ public boolean isEmpty() {
...
+            final boolean failOnLimits =
+                options.getFailOnLimits().orElse(LocalMetricManager.this.failOnLimits);
+
+            final OptionalLimit groupLimit =
+                options.getGroupLimit().orElse(LocalMetricManager.this.groupLimit);
+
+            // Transform that takes the result from ES metadata lookup to fetch from backend
+            final LazyTransform<FindSeries, FullQuery> transform =
+                new Transform(request, failOnLimits, seriesLimit, groupLimit, quotaWatcher,
+                              dataInMemoryReporter);
```
gabrielgerhardsson: 👍  Thanks

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -481,7 +521,7 @@ public FullQuery end(int resolved, int failed, int cancelled) throws Exception {
                     checkIssues(failed, cancelled).orElse("Query exceeded quota")));
 
                 return new FullQuery(trace, errorsBuilder.build(), ImmutableList.of(),
-                    Statistics.empty(), new ResultLimits(limitsBuilder.build()));
+                                     Statistics.empty(), new ResultLimits(limitsBuilder.build()));
```
gabrielgerhardsson: There's still a few formatting changes like this that would be cancelled out if you auto-format. So please do apply the formatter before committing.

udoprog: Could possibly using the wrong style config?
https://github.com/spotify/heroic/tree/master/idea

PR #287, opened 2017-05-30
[datastax] Upgrade & repackage Datastax driver

gabrielgerhardsson:
NOTE: You need to re-run ./tools/install-repackaged after this commit.

* Upgrades Datastax driver to 3.2.0
* Isolates the Datastax driver so that its dependencies doesn't clash
with other Heroic dependencies.

### DIFF
--- a/metric/datastax/pom.xml
+++ b/metric/datastax/pom.xml
@@ -16,6 +16,10 @@
 
   <name>Heroic: Datastax Backend</name>
 
+  <properties>
+    <datastax.version>3.2.0</datastax.version>
+  </properties>
+
   <dependencies>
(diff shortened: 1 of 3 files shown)


gabrielgerhardsson:
Integration tests pass, with Cassandra 3.5.

udoprog:
@gabrielgerhardsson cool!

But we need to test it with 2.1.x as well since it's listed in the readme.

gabrielgerhardsson:
@udoprog Ah right, I'm on it.

gabrielgerhardsson:
Now confirmed with Cassandra 2.1 also :)
Merging.

PR #290, opened 2017-06-02
Better min/max statistics for high-rate Histograms

gabrielgerhardsson:
The common semantic metrics Histogram uses a reservoir that provides statistically relevant data without saving all samples, by performing random sampling. That, however, means that min/max isn't accurate for high-rate metrics.
This commit introduces a new reservoir that can be layered on top of the old one, to keep special track of min/max values.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -737,15 +737,15 @@
 
       <!-- testing -->
       <dependency>
-        <groupId>junit</groupId>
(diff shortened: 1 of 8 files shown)


udoprog:
@gabrielgerhardsson merge conflicts, otherwise it lgtm

PR #291, opened 2017-06-02
[elasticsearch] Slow-start mode for rate limiting

gabrielgerhardsson:
Existing functionality: The option writesPerSecond enables a rate limit
of how many writes that may be performed per second. Any additional
write requests will be dropped.
New: This commit adds a possibility to have a slow-start mode for this
rate limit, causing the limit to be 3 times lower for the x first
seconds.
Disabled by deafult. Controlled by the new rateLimitSlowStartSeconds
option for metadata and suggest.

### DIFF
--- a/metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/ElasticsearchMetadataModule.java
+++ b/metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/ElasticsearchMetadataModule.java
@@ -21,6 +21,11 @@
 
 package com.spotify.heroic.metadata.elasticsearch;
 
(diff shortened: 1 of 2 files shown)


udoprog:
@gabrielgerhardsson you might have forgotten to commit changes to `RateLimiter#create(..)`

gabrielgerhardsson:
@udoprog You're fast! :)
Hmm, no my intention was to use the existing "public static RateLimiter create(double permitsPerSecond, long warmupPeriod, TimeUnit unit)". Do you see any issues?

I see that my branch was also outdated. Will fix.

udoprog:
Oh, my bad. I thought `RateLimiter` was an internal type. Carry on :).

PR #292, opened 2017-06-02
[aggregation] sum squared

mykolasmith:
I noticed that the `spread` aggregator includes `sum2` (sum of squares), which we are using for standard deviation calculations. I am proposing that `sum2` be made into a first-class aggregation (along with min, max, count, sum, etc.) It's a small, isolated change. Tested and working in our QA cluster.

### DIFF
--- a/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/Sum2Bucket.java
+++ b/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/Sum2Bucket.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright (c) 2015 Spotify AB.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.heroic.aggregation.simple;
+
+import com.google.common.util.concurrent.AtomicDouble;
+import com.spotify.heroic.aggregation.AbstractBucket;
+import com.spotify.heroic.aggregation.DoubleBucket;
+import com.spotify.heroic.metric.Point;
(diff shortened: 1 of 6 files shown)


aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/Sum2Bucket.java
```diff
@@ -0,0 +1,75 @@
...
+    }
+
+    @Override
+    public double value() {
+        if (!valid) {
+            return Double.NaN;
+}
+
+        return sum2.get();
+            }
```
udoprog: indentation issue here?

aggregation/simple/src/test/java/com/spotify/heroic/aggregation/simple/Sum2BucketTest.java
```diff
@@ -0,0 +1,37 @@
...
+            assertTrue(bucket.getClass().getSimpleName(), Double.isNaN(bucket.value()));
+        }
+    }
+
+    @Test
+    public void testAddSome() {
+        for (final DoubleBucket bucket : buckets()) {
+            bucket.updatePoint(tags, new Point(0, 10.0));
+            bucket.updatePoint(tags, new Point(0, 20.0));
+            assertEquals(bucket.getClass().getSimpleName(), 500.0, bucket.value(), 0.0);
```
udoprog: I think this would generate a bit of an odd error message.

The epsilon value should preferably be non-zero. 0.1-0.5 seems like a decent value.

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/Sum2Bucket.java
udoprog: no idea why this isn't caught by checkstyle:
https://github.com/spotify/heroic/issues/293

aggregation/simple/src/test/java/com/spotify/heroic/aggregation/simple/Sum2BucketTest.java
mykolasmith: Fixed the formatting issue. Not sure what you mean, this shouldn't generated any error message. I pretty much copy/pasted the SumBucketTest and updated the expected result. If you're saying I should make the TS different for the second point, e.g. `new Point(0.1, 20.0)`, would you like me to also update the SumBucketTest?

mykolasmith: https://github.com/spotify/heroic/blob/master/aggregation/simple/src/test/java/com/spotify/heroic/aggregation/simple/SumBucketTest.java#L31-L34

udoprog: For the epsilon: the last argument is the permitted difference from the expected value. It should probably be increased for both test cases to be non-zero, but since it effectively works now it's well-defined enough. This could become an issue if someone else starts fiddling with them.

Because this is a copy-paste, I'm OK with it. We'll try to fix both in a future patch.

udoprog:
Thanks!

PR #303, opened 2017-06-26
[metric/bigtable] Ensure correct statistics

gabrielgerhardsson:
Enforce execution order so that rowsAccessed statistic is always updated
before being read.


zalenski:
👍

PR #305, opened 2017-06-28
Introduce memcached-backed caching using folsom

udoprog:


heroic-core/src/main/java/com/spotify/heroic/CoreQueryManager.java
```diff
@@ -280,14 +280,15 @@ public Aggregation visitString(final StringExpression e) {
...
-                        return result;
-                    })
-                    .onDone(onDoneQueryReporter);
+                final AsyncFuture<QueryResult> collect =
+                    async.collect(futures, QueryResult.collectParts(QUERY, range, combiner, limit));
+                return collect;
             });
+
+            return query.directTransform(result -> {
+                reportCompletedQuery(result, fullQueryWatch);
```
gabrielgerhardsson: Would it make sense to somehow flag in the query log whether a response was cached or not? It would seem nice to be able to ignore cached queries when doing certain analysis of the query logs.

udoprog: Yeah, sure.

PR #306, opened 2017-06-28
[component, core, dist] Prepare for darkload

gabrielgerhardsson:
Do rpc protocol breaking change now, in preparation for darkload
feature.

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/cluster/NodeMetadata.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/cluster/NodeMetadata.java
@@ -24,10 +24,9 @@ package com.spotify.heroic.cluster;
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.spotify.heroic.common.ServiceInfo;
-import lombok.Data;
-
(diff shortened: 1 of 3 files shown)


heroic-component/src/main/java/com/spotify/heroic/cluster/NodeMetadata.java
```diff
@@ -24,27 +24,29 @@
...
     private final Map<String, String> tags;
     private final ServiceInfo service;
+    private final Boolean darkload;
 
     @JsonCreator
     public NodeMetadata(
         @JsonProperty("version") Integer version, @JsonProperty("id") UUID id,
-        @JsonProperty("tags") Map<String, String> tags, @JsonProperty("service") ServiceInfo service
+        @JsonProperty("tags") Map<String, String> tags,
+        @JsonProperty("service") ServiceInfo service, @JsonProperty("darkload") Boolean darkload
```
udoprog: use primitive

gabrielgerhardsson: Ok, fixed

udoprog:
@gabrielgerhardsson came up with an alternative solution that doesn't require a protocol change.

PR #308, opened 2017-06-28
[component,core] Add darkload functionality

gabrielgerhardsson:
When testing a new configuration option, backend type, code change, etc
it can be very good to be able to test with the same query load as your
main production cluster. This commit adds this possibility.
Set magic tag darkload:true on your shard. This will cause queries to
fan out to it, but the result will be discarded.

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterShard.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterShard.java
@@ -45,6 +45,7 @@ import lombok.extern.slf4j.Slf4j;
 public class ClusterShard {
     private static final QueryTrace.Identifier RETRY_BACKOFF =
         new QueryTrace.Identifier("retry-backoff");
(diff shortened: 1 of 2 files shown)


heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterShard.java
```diff
@@ -115,6 +116,18 @@
         return nodes;
     }
 
+    public boolean isDarkload() {
+        if (!shard.containsKey(DARKLOAD)) {
```
udoprog: Only this check might be enough?

gabrielgerhardsson: Sure 👍  I'll fix it

gabrielgerhardsson: Fixed

PR #317, opened 2017-07-04
[metadata,suggest] Metrics for actual writes

gabrielgerhardsson:
Renamed write-* group of metrics to cached-write-* for metadata and
suggest. These metrics measure the full set of writes, including the
ones that hit the internal write-cache in these modules, and thus
actually aren't sent to ES.

Added new write-* group of metrics. These measure the actual writes sent
to ES.

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/statistics/MetadataBackendReporter.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/statistics/MetadataBackendReporter.java
@@ -26,6 +26,8 @@ import com.spotify.heroic.metadata.MetadataBackend;
 public interface MetadataBackendReporter {
     void reportWriteDroppedByRateLimit();
 
+    FutureReporter.Context setupWriteReporter();
+
     /**
      * report number of successful operations in a batch
      *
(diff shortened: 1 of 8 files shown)


udoprog:
@gabrielgerhardsson not sure `raw` is the best terminology here. It's not descriptive of the circumstance.

If the write is already distinguished by some other tag (possibly component?), then we can re-use the simple name (`write`). Otherwise, I would say the cached variant is the special case that should be prefixed so that it can be distinguished somehow by the principle of least astonishment. Writes are expected to be towards the backend in question.

gabrielgerhardsson:
@udoprog Very reasonable comment. noncached felt a tad long. But if we switch things around as you suggest then it feels more realistic. So suggestion:
* write-cached-resolve-rate - all writes, including the ones hitting the write-cache and never actually reaching ES. Not super happy with "-cached" tho since this is not only cached writes, it's all writes. Hmm.
* write-resolve-rate - only the writes that actually is sent to ES.

udoprog:
`noncached` is actually not bad, it's descriptive at least. if it turns out that naming it the other way is complicated, that seems like a good alternative.

gabrielgerhardsson:
@udoprog Yes noncached is descriptive indeed. But here's a try at making write-* actually represent the writes going to ES, and adding a new cached- prefix for the current write-* metrics.

PR #318, opened 2017-07-05
[component] Log at rpc failure & cancel

gabrielgerhardsson:
Log when there is a failure or cancellation in an RPC call. The goal is
to give insights into the different circumstances where this happens.

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterShard.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterShard.java
@@ -90,12 +90,16 @@ public class ClusterShard {
 
                 return result.getReturnValue().catchFailed(throwable -> {
                     reporter.reportClusterNodeRpcError();
+                    log.info(
+                        "Cluster RPC call failed for node " + result.getNode().toString() + ": " +
+                            throwable.getMessage());
                     /* Actually never return;s, instead throws a new exception with added info.
                      * The point is to get Node identifying information into the exception */
                     throw new RuntimeNodeException(result.getNode().toString(),
                         throwable.getMessage(), throwable);
                 }).catchCancelled(ignore -> {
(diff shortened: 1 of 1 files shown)


heroic-component/src/main/java/com/spotify/heroic/cluster/ClusterShard.java
```diff
@@ -90,12 +90,16 @@
 
                 return result.getReturnValue().catchFailed(throwable -> {
                     reporter.reportClusterNodeRpcError();
+                    log.info(
+                        "Cluster RPC call failed for node " + result.getNode().toString() + ": " +
```
udoprog: Prefer string formatting with `{}`:
https://www.slf4j.org/faq.html#logging_performance

udoprog:
Not sure how useful these errors will be. They are expected during restarts and would become quite spammy while not actually contributing to troubleshooting. And it's already instrumented as a metric.

The actual error message will always be something of the kind (when using gRPC): `Request finished with status code (DEADLINE_EXCEEDED)`.

Perhaps use something that logs all captured errors in the case of complete shard failure instead, preferably all errors should be propagated anyways to a downstream consumer as suppressed exceptions.

udoprog:
Alternatively, make this a debug or trace log level, so it can be disabled. Unless there is a complete shard failure, in which case it should be propagated and logged downstream.

udoprog:
It's also worth noting that we already log the complete exception server-side:
https://github.com/spotify/heroic/blob/master/rpc/grpc/src/main/java/com/spotify/heroic/rpc/grpc/GrpcRpcProtocolServer.java#L232

gabrielgerhardsson:
@udoprog Thanks. Your last comment makes me believe that this whole PR should be tossed.
As to exceptions being propagated, they already are propagated and shown in the query response as suppressed errors in the trace.

PR #320, opened 2017-07-05
[component,core] more detailed caching information

udoprog:
This introduces the `CacheInfo` object as a part of the `QueryMetricsResponse`.

This tells us:
* What the cache key used was.
* What the ttl of the cached item was.
* If a cached result was used, or not.

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/metric/CacheInfo.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/metric/CacheInfo.java
@@ -1,0 +1,42 @@
+/*
+ * Copyright (c) 2015 Spotify AB.
+ *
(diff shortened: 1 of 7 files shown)


heroic-component/src/main/java/com/spotify/heroic/metric/CacheInfo.java
```diff
@@ -0,0 +1,42 @@
+/*
+ * Copyright (c) 2015 Spotify AB.
```
gabrielgerhardsson: 2017 :)

udoprog: Hah, I am happily going to ignore this. Thanks.

PR #325, opened 2017-07-13
[metadata,suggest] Rework write metrics

gabrielgerhardsson:
write-* - the full set of write requests coming to metadata/suggest BE
writes-dropped-by-cache-hit - writes dropped by the write cache
backend-write-* - the write requests that are actually sent to ES
writes-dropped-by-rate-limit - write requests where ES says the document
already existed.

### DIFF
--- a/metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
+++ b/metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
@@ -87,6 +87,7 @@ import java.util.stream.Collectors;
 import javax.inject.Inject;
 import javax.inject.Named;
 import lombok.ToString;
+import lombok.extern.slf4j.Slf4j;
 import org.apache.commons.lang3.tuple.Pair;
 import org.elasticsearch.action.ActionRequestBuilder;
 import org.elasticsearch.action.DocWriteRequest.OpType;
@@ -105,6 +106,7 @@ import org.elasticsearch.search.aggregations.AggregationBuilder;
 import org.elasticsearch.search.aggregations.AggregationBuilders;
 import org.elasticsearch.search.aggregations.bucket.terms.Terms;
 
+@Slf4j
 @ElasticsearchScope
 @ToString(of = {"connection"})
 public class MetadataBackendKV extends AbstractElasticsearchMetadataBackend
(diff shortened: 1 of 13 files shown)


suggest/elasticsearch/src/main/java/com/spotify/heroic/suggest/elasticsearch/SuggestBackendKV.java
```diff
@@ -578,7 +581,19 @@ public boolean isReady() {
...
+                        .catchFailed(throwable -> {
+                            if (throwable instanceof VersionConflictEngineException) {
+                                /* Index request rejected, document already exists. That's ok,
+                                 * return success. */
+                                reporter.reportWriteDroppedByDuplicate();
+                                return WriteSuggest.of();
+                            }
+                            log.error("Suggest write failed: ", throwable);
+                            throw new RuntimeException(throwable);
+                        })
```
udoprog: Pull this to a helper method?

```
catchFailed(this::handleVersionConflict)
```

Logging should not be necessary. If it's needed, it should happen downstream since the exception is propagated.

metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
```diff
@@ -206,10 +210,20 @@ public Groups groups() {
...
+                    .catchFailed(throwable -> {
+                        if (throwable instanceof VersionConflictEngineException) {
+                            /* Index request rejected, document already exists. That's ok, return
+                             * success. */
+                            reporter.reportWriteDroppedByDuplicate();
+                            return WriteMetadata.of();
+                        }
+                        log.error("Metadata write failed: ", throwable);
+                        throw new RuntimeException(throwable);
+                    })
```
udoprog: Use helper method.

suggest/elasticsearch/src/main/java/com/spotify/heroic/suggest/elasticsearch/SuggestBackendKV.java
gabrielgerhardsson: 👍

metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
gabrielgerhardsson: 👍

gabrielgerhardsson:
@udoprog Thanks for the review. I've fixed those two things now.

udoprog:
Works for me, but I also think that the two Elasticsearch backends have a common base class that could host the `handleVersionConflict` method.

gabrielgerhardsson:
@udoprog Good point, fixed now.

PR #326, opened 2017-07-13
Fix consumer slow-start mode

gabrielgerhardsson:
Reimplement so that we avoid using a seemingly problematic rate limiter implementation.

Instead of using the Guava SmoothWarmingUp implementation, use the
simple SmoothBursty and apply a manual slow start period to it in a
wrapper. Slow start duration is still configurable with
rateLimitSlowStartSeconds. The amount of slow start is hard coded to
factor 4. During startup the rate will be 4 times lower for x seconds.

### DIFF
--- a/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/SlowStartRateLimiter.java
+++ b/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/SlowStartRateLimiter.java
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2015 Spotify AB.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.heroic.elasticsearch;
+
+import com.google.common.util.concurrent.RateLimiter;
+import com.spotify.heroic.time.Clock;
+import java.util.concurrent.TimeUnit;
+import lombok.Data;
+
+@Data
+public class SlowStartRateLimiter implements SimpleRateLimiter {
+    private final RateLimiter rateLimiter;
+    private final double ratePerSecond;
+    private final long slowStartDuration;
+    private final TimeUnit slowStartTimeUnit;
+    private final Clock clock;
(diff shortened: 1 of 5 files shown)


heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/SlowStartRateLimiter.java
```diff
@@ -0,0 +1,81 @@
...
+    private final long slowStartDuration;
+    private final TimeUnit slowStartTimeUnit;
+    private final Clock clock;
+    private final long startTime;
+
+    /* Don't use an atomic type, to improve performance. The cost of this is that during startup
+     * there might be some extra slow path calls verifying the time in threads until all threads
+     * has seen the updated boolean value. No harm tho. This only affects startup performance and
+     * improves performance after. Repeated calls to setRate() are ok. */
+    private boolean isSlowStartMode;
```
udoprog: nice comment, but needs to be volatile though.
Otherwise the update might not ever be visible to all threads.

gabrielgerhardsson: Right! Ha, true story, I actually thought of this needing to be volatile a couple of nights ago but forgot to write it down. Good catch.

gabrielgerhardsson:
@udoprog Can we revisit this? I rebased again and pushed now.

udoprog:
@gabrielgerhardsson hm, it might be more straight forward if just set a timer at start (using the global `ScheduledExecutorService`) that will update the rate accordingly instead of spinning on `isSlowStartMode` and then racing to call `setRate`.

gabrielgerhardsson:
@udoprog That's a good point, thanks. That would be much more straight forward.
I'll look at this when I get the chance.

dmichel1:
@gabrielgerhardsson do you think this is still needed? Do you remember what you were observing internally with the rate limiter implementation?

gabrielgerhardsson:
@dmichel1 The were observing spikes of requests going out from consumer nodes to ES clusters, after restart (and after reconnect, if I remember correctly). We believed it was caused by the SmoothWarmingUp rate limiter - looked like it was letting too much traffic through at the very beginning, which was drowning the ES clusters in traffic when they first came back up after being down. Possibly, the ES clusters were extra sensitive to load at the very beginning, perhaps that they were still syncing the shards among nodes etc. So we felt it important to have a proper slow start method.
We decided to try a more simple one instead. The more simple one did look good indeed. 

I'm not able to verify right now that the current code uses the same SmoothWarmingUp. I seem to remember that we switched to a straight rate limit instead (and probably choose a lower rate value).

Using the warmup implementation in this ticket could possibly allow us to increase the write rate to ES per node, while maintaining a slow rate speed at warmup, making us not drown ES. Worth looking into perhaps.

dmichel1:
@gabrielgerhardsson thank you for the explanation.

I verified the rate limiter is still using the same `SmoothWarmingUp` even after the upgrade to guava 20 (from 19).

One thing we observed when migrating to Google Pubsub is the cache expiration happening roughly at the same time across all consumers (because they are often restarted or deployed together). This puts a LOT of extra load on the ES clusters until the write-though-cache fills up again.  

screen shot 2018-09-24 at 1 51 49 pm

We mostly smoothed out this graph with the new consumers by adding a random amount of time to each one.

So we have not specially ran in the issue you are describing but this is very useful information for us to know so we can watch out for it.

I'm going to close this and we can re-open if this comes up again.

Thanks,
Drew

PR #330, opened 2017-08-14
[all] provide object-based deserialization for filters

udoprog:
This commit adds object-based deserialization for filters.

Object-based deserialization emulates Jackson's type-based serialization, by allowing the following kind of objects to be accepted as filters (these two are now equivalent):
```json
["=", "foo", "bar"]
{"type": "matchTag", "tag": "foo", "value": "bar"}
```
The reason behind doing this is to better support static languages, and their corresponding serialization frameworks. While it is very common to immediately support the second variant, the first one is not.

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/filter/MatchKeyFilter.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/filter/MatchKeyFilter.java
@@ -21,6 +21,7 @@
 
 package com.spotify.heroic.filter;
 
+import com.fasterxml.jackson.annotation.JsonTypeName;
 import com.spotify.heroic.ObjectHasher;
 import com.spotify.heroic.common.Series;
 import com.spotify.heroic.grammar.DSL;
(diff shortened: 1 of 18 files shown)


heroic-component/src/main/java/com/spotify/heroic/filter/MatchKeyFilter.java
```diff
@@ -28,15 +28,15 @@
 import lombok.EqualsAndHashCode;
 
 @Data
-@EqualsAndHashCode(of = {"OPERATOR", "value"}, doNotUseGetters = true)
+@EqualsAndHashCode(of = {"OPERATOR", "key"}, doNotUseGetters = true)
 public class MatchKeyFilter implements Filter {
```
gabrielgerhardsson: Is there a specific reason for having @JsonTypeName specified for some filters but not all?

udoprog: No. It's better to include it. It worked without `@JsonTypeName` because it defaulted to the operator, which for key is `key`.

Fixed, thanks!

gabrielgerhardsson: Ok thanks for the explanation. LGTM!

PR #334, opened 2017-09-21
Clean up MetricCollection

udoprog:
This converts up the current `abstract` MetricCollection into an interface, and uses a standard serialization method in fasterxml instead of the custom serializer (see removal of `MetricCollectionSerialization`).

This is in preparation of experimenting with additional Metric types, in which modifying MetricCollection would be a central change. Simplifying and streamlining this will improve the clarity of future pull requests.

While doing this, a few bugs to existing Metrics were detected, and tests introduced in `BasicSerialization`.

The `EmptyMetricCollection` null object has been removed. Attempting to share it between nodes results in inconsistent behavior, which is just confusing.


heroic-component/src/test/java/com/spotify/heroic/aggregation/AggregationOutputTest.java
```diff
@@ -20,16 +20,16 @@ public void lombokDataTest() {
 
     @Test
     public void isEmpty() {
-        final AggregationOutput output =
-            new AggregationOutput(ImmutableMap.of(), ImmutableSet.of(), MetricCollection.empty());
+        final AggregationOutput output = new AggregationOutput(ImmutableMap.of(), ImmutableSet.of(),
+            MetricCollection.points(ImmutableList.of()));
```
gabrielgerhardsson: Say, what's the reason behind removing the EmptyMetricCollection class?

heroic-core/src/main/java/com/spotify/heroic/http/write/WriteResource.java
```diff
@@ -52,7 +60,13 @@ public void metrics(
         @Suspended final AsyncResponse response, @QueryParam("group") String group,
         WriteMetricRequest write
     ) throws Exception {
-        httpAsync.bind(response, ingestion.useGroup(group).write(write.toIngestionRequest()),
-            r -> r);
+        final IngestionGroup ingestionGroup = ingestion.useGroup(group);
```
gabrielgerhardsson: Unrelated change to this commit?

heroic-component/src/test/java/com/spotify/heroic/aggregation/AggregationOutputTest.java
udoprog: Unnecessary null object. It's also mentioned in the description:

> The EmptyMetricCollection null object has been removed. Attempting to share it between nodes results in inconsistent behavior, which is just confusing.

heroic-core/src/main/java/com/spotify/heroic/http/write/WriteResource.java
udoprog: This happens because `EmptyMetricCollection` was removed in favor of `Optional<MetricCollection>`

heroic-component/src/test/java/com/spotify/heroic/aggregation/AggregationOutputTest.java
gabrielgerhardsson: Ok I see, thanks. Sorry, didn't see the description note regarding that.

heroic-core/src/main/java/com/spotify/heroic/http/write/WriteResource.java
gabrielgerhardsson: Ok 👍

PR #338, opened 2017-10-03
[core] new task annotated-export

udoprog:
This task export each time series (matching a filter) into one file per time series (`<hash>.json`).

The first line of the file is the time series, the rest are the samples exported.

```json
{"key":"foo","tags":{"what":"hello-world","unit":"s"}}
[1234567890123,42.0]
[1234567890124,42.0]
```

This task is not perfect (yet):
 * The type of data exported should be configurable, it's currently hardcoded to `point`. There is also a typed call to `getDataAs(...)` that will break unless points are being exported.
* Output is _not_ limited. Use at your own risk.
* Doesn't fully use the new streaming APIs, so memory use is determined by the maximum amount of data exported for a given time series.

### DIFF
--- a/heroic-core/src/main/java/com/spotify/heroic/shell/task/AnnotatedExport.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/shell/task/AnnotatedExport.java
@@ -1,0 +1,260 @@
+/*
+ * Copyright (c) 2015 Spotify AB.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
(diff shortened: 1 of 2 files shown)


heroic-core/src/main/java/com/spotify/heroic/shell/task/AnnotatedExport.java
```diff
@@ -0,0 +1,260 @@
...
+                public void end() {
+                    final List<Callable<AsyncFuture<SeriesAndData>>> callables = new ArrayList<>();
+
+                    for (final Series s : allSeries) {
+                        callables.add(() -> {
+                            final FetchData.Request request =
+                                new FetchData.Request(MetricType.POINT, s, range,
+                                    QueryOptions.defaults());
+                            return m
+                                .fetch(request, FetchQuotaWatcher.NO_QUOTA)
```
gabrielgerhardsson: This uses the deprecated variant of fetch(), so non-streaming. As long as the data fits in memory that's not a big problem, but if we want to export more then it would have been great to do it in a streaming way. How about adding a TODO or fix?

heroic-core/src/main/java/com/spotify/heroic/shell/task/AnnotatedExport.java
```diff
@@ -0,0 +1,260 @@
...
+                        });
+                }
+            });
+
+        return future;
+    }
+
+    @ToString
+    private static class Parameters extends Tasks.QueryParamsBase {
+        @Option(name = "-f", aliases = {"--file"}, usage = "File to perform test against",
```
gabrielgerhardsson: Unused param?

hexedpackets:
Closing due to age, feel free to re-open if this is still an issue.

PR #341, opened 2017-11-08
[aggregation] Fix Delta aggregation with dist agg

gabrielgerhardsson:
When using distributed aggregations, a delta aggregation were combining the data with a second delta aggregation, causing inaccurate result.


gabrielgerhardsson:
FYI: Not yet ready to be merged.

PR #342, opened 2017-11-09
Back to bt 0.9.6.2

gabrielgerhardsson:
Revert the BT client library from 0.9.7.1 to 0.9.6.2, because of a known issue in 0.9.7.1. Version 1.0.0 will have the fix, but is not out yet.

### DIFF
--- a/metric/bigtable/pom.xml
+++ b/metric/bigtable/pom.xml
@@ -22,7 +22,7 @@
   </description>
 
   <properties>
-    <bigtable.version>0.9.7.1</bigtable.version>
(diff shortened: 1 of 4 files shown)


whitesyn:
👍

nikek:
👍

gabrielgerhardsson:
FYI, the issue with 0.9.7.1:
This version exhibited an issue where read and write operations could get stuck indefinitely sometimes. Heroic has an internal limit on the number of outstanding write requests, so when it hit that no more write requests would be allowed through, which stopped writes altogether.

PR #344, opened 2017-11-27
[core,component] allow limiting concurrent queries

gabrielgerhardsson:
Limits how many concurrent queries that are allowed on a data-node
level. When the limit is reached, backoff is applied and the cluster
will retry with another node in that shard.
concurrentQueriesBackoff in the metric section configures this limit.
Unlimited by default.


heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -296,6 +303,13 @@ public QueryTrace buildTrace() {
 
         @Override
         public AsyncFuture<FullQuery> query(final FullQuery.Request request) {
+            if (!concurrentQueries.tryAcquire()) {
+                // There's currently too many concurrent queries. Fail now so that the QueryManager
+                // gets an opportunity to try another node in the same shard instead.
+                throw new GoAwayException(
+                    "Node has reached maximum number of concurrent MetricManager requests (" +
+                        concurrentQueriesBackoff + ")");
+            }
```
udoprog: Any synchronous exception in this code block will cause the semaphore to leak leases, and eventually lock up.

I'd suggest using a helper function that catches these, and releases the semaphore for but circumstances.
* An immediate exception is thrown.
* The future completes.

gabrielgerhardsson: Can't believe I missed that 😊  Thanks

gabrielgerhardsson: Fixed this now, in addition to spotting an issue that existed previously also: A non thread safe list being used for keeping track of which nodes that had been tried. Introduced by me when working on retry code some time ago. Fixed now :)

gabrielgerhardsson:
@udoprog Pushed an update. Could I please get another code review?

udoprog:
Cool! Will have a look when I've managed to get more sleep.

gabrielgerhardsson:
No rush at all! @udoprog

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -296,7 +303,24 @@ public QueryTrace buildTrace() {
...
+                // gets an opportunity to try another node in the same shard instead.
+                return async.failed(new GoAwayException(
+                    "Node has reached maximum number of concurrent MetricManager requests (" +
+                        concurrentQueriesBackoff + ")"));
+            }
+
+            try {
+                return protectedQuery(request).onFinished(concurrentQueries::release);
+            } catch (Exception e) {
+                log.info("Failed, ensuring that we release the semaphore");
```
udoprog: This log entry is not needed IMO.

PR #348, opened 2017-12-13
Bt 1 0 0

nezdolik:


metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/api/BigtableMutatorImpl.java
```diff
@@ -166,7 +156,14 @@ public void onFailure(Throwable t) {
 
     private void flush() {
         synchronized (lock) {
-            tableToBulkMutation.values().stream().forEach(BulkMutation::flush);
+            tableToBulkMutation.values().stream().forEach(mutation -> {
+                try {
+                    mutation.flush();
+                } catch (final InterruptedException e) {
+                    log.error("Failed to apply mutation {}:{}", mutation, e.getMessage());
```
udoprog: Don't log here. Exception is re-thrown with all relevant information and mutations could be verbose.

PR #352, opened 2017-12-23
Fix serialization of pointsabove and pointsbelow

udoprog:
@mykolasmith could you take this for a spin and see if it fixes your issue?


udoprog:
@mykolasmith could you take this for a spin and see if it fixes your issue?

Issue #351, opened 2017-12-23
Frequent JSON Mapping Exceptions


mykolasmith:
Hi, we're seeing lots of JSON Mapping Exceptions. Can you take a brief look at the PointsAboveInstance and PointsBelowInstance and see if they are implemented correctly? Thanks...

```
2017-12-23 14:40:36,318 ERROR c.s.h.r.g.GrpcRpcProtocolServer [grpc-default-executor-20] 4725ed4d-85c2-47de-843a-28689473dd85: Failed to handle request (sent Status{code=INTERNAL, description=null, cause=null}) com.fasterxml.jackson.databind.JsonMappingException: No suitable constructor found for type [simple type, class com.spotify.heroic.aggregation.simple.PointsAboveInstance]: can not instantiate from JSON object (missing default constructor or creator, or perhaps need to add/enable type information?)

 at [Source: [B@4cf0d6f4; line: 1, column: 254] (through reference chain: com.spotify.heroic.rpc.grpc.GroupedQuery["query"]->com.spotify.heroic.metric.Request["aggregation"]->com.spotify.heroic.aggregation.GroupInstance["each"]->com.spotify.heroic.aggregation.ChainInstance["chain"]->java.util.ArrayList[1])
```

udoprog:
Thanks for the report!

Assuming you are running the latest commit?

Yeah, this looks like a bug. We don't seem to have any integration tests which include these aggregations, that would help to nail down the problem.

mykolasmith:
@udoprog ahhh, looks like you beat me to it! I've just tested this PR on our QA cluster and have confirmed that it fixes the issue. https://github.com/spotify/heroic/pull/353

udoprog:
@mykolasmith incorporate my test-case into your fix and I'll close mine and merge yours :).

mykolasmith:
@udoprog done & passing 👍

PR #353, opened 2017-12-23
[aggregation] Fix PointsAbove/PointsBelow JSON constructor

mykolasmith:
Sets the proper JSONCreator properties on the PointsAbove and PointsBelow instances, which addresses https://github.com/spotify/heroic/issues/351.

### DIFF
--- a/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/MetricMappingAggregation.java
+++ b/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/MetricMappingAggregation.java
@@ -23,6 +23,7 @@ package com.spotify.heroic.aggregation.simple;
 
 import static java.util.stream.Collectors.toList;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.spotify.heroic.ObjectHasher;
 import com.spotify.heroic.aggregation.AggregationInstance;
 import com.spotify.heroic.aggregation.AggregationOutput;
@@ -41,10 +42,15 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import lombok.Data;
+import lombok.NonNull;
 
 @Data
 public abstract class MetricMappingAggregation implements AggregationInstance {
+
     private static final EmptyInstance INNER = EmptyInstance.INSTANCE;
+
+    @NonNull
+    @JsonIgnore
     private final MetricMappingStrategy metricMappingStrategy;
 
     @Override
--- a/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/PointsAboveInstance.java
+++ b/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/PointsAboveInstance.java
@@ -22,16 +22,18 @@
 package com.spotify.heroic.aggregation.simple;
 
 import lombok.EqualsAndHashCode;
-import lombok.ToString;
+import lombok.Data;
 
 import java.beans.ConstructorProperties;
 
-@ToString
+@Data
 @EqualsAndHashCode(callSuper = true)
 public class PointsAboveInstance extends MetricMappingAggregation {
+    private final double threshold;
 
     @ConstructorProperties({"threshold"})
-    public PointsAboveInstance(double threshold) {
+    public PointsAboveInstance(final double threshold) {
         super(new FilterPointsThresholdStrategy(FilterKThresholdType.ABOVE, threshold));
+        this.threshold = threshold;
     }
 }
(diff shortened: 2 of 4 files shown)


aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/PointsAboveInstance.java
```diff
@@ -21,17 +21,19 @@
...
 
-import java.beans.ConstructorProperties;
-
-@ToString
+@Data
 @EqualsAndHashCode(callSuper = true)
 public class PointsAboveInstance extends MetricMappingAggregation {
+    private final double threshold;
 
-    @ConstructorProperties({"threshold"})
```
udoprog: So we actually want to make use of `ConstructorProperties` here. The problem was (I think) that mapping was being serialized which you have already prevented. So this change should not be necessary.

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/PointsBelowInstance.java
```diff
@@ -21,11 +21,19 @@
...
-import java.beans.ConstructorProperties;
+import com.fasterxml.jackson.annotation.JsonCreator;
+import com.fasterxml.jackson.annotation.JsonProperty;
+import lombok.Data;
+import lombok.EqualsAndHashCode;
 
+@Data
+@EqualsAndHashCode(callSuper = true)
 public class PointsBelowInstance extends MetricMappingAggregation {
-    @ConstructorProperties({"threshold"})
```
udoprog: Same here.

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/MetricMappingAggregation.java
```diff
@@ -41,10 +42,15 @@
...
 import java.util.Set;
 import lombok.Data;
+import lombok.NonNull;
 
 @Data
 public abstract class MetricMappingAggregation implements AggregationInstance {
+
     private static final EmptyInstance INNER = EmptyInstance.INSTANCE;
+
+    @NonNull
```
udoprog: `NonNull` is not needed?

mykolasmith: Copied from here: https://github.com/OdenTech/heroic/blob/fix-filter-points-agg/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/FilterAggregation.java#L52

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/PointsAboveInstance.java
mykolasmith: I didn't see it called on the AboveKInstance. Just trying to be consistent. I'm happy to add it back in, just not too sure what it does.

udoprog: Yeah, please do. Using `ConstructorProperties` is a standard annotation and is currently preferred. We'll probably do a small project getting rid of the `JsonCreator` et. al. in the future.

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/MetricMappingAggregation.java
udoprog: Yeah OK. It's mostly harmless so I'm OK with it. Here it would generate a null-check in the constructor which we (should) already disallow in serialization.

mykolasmith:
Made requested changes and squashed all into single commit.

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/PointsBelowInstance.java
```diff
@@ -21,11 +21,22 @@
...
+import lombok.EqualsAndHashCode;
+
 import java.beans.ConstructorProperties;
 
+@Data
+@EqualsAndHashCode(callSuper = true)
 public class PointsBelowInstance extends MetricMappingAggregation {
+    private final double threshold;
+
+    @JsonCreator
```
udoprog: `@ConstructorProperties` completely replaces `@JsonCreator`. Therefore `@JsonCreator` and `@JsonProperty` annotations should not be present if `@ConstructorProperties` is present.

Apologies if I didn't make that clear enough.

mykolasmith: I'm pretty sure that's what fixed this issue in the first place? Hang on, going to test it now without JSONCreatory/Property.

mykolasmith: Ok, all set.

udoprog:
Thanks!

PR #354, opened 2018-01-05
Remove non-sliced metric fetch code path

gabrielgerhardsson:
This code path was already marked deprecated before, and was kept
as a safety fallback while adding the sliced fetch support.
Removing this code will make it easier to do future changes.

### DIFF
--- a/heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
@@ -176,7 +176,6 @@ public class LocalMetricManager implements MetricManager {
             private final QueryOptions options;
             private final DataInMemoryReporter dataInMemoryReporter;
             private final MetricType source;
-            private final boolean slicedFetch;
 
             private Transform(
                 final FullQuery.Request request, final boolean failOnLimits,
@@ -198,7 +197,6 @@ public class LocalMetricManager implements MetricManager {
                 this.dataInMemoryReporter = dataInMemoryReporter;
 
                 final Features features = request.getFeatures();
-                this.slicedFetch = features.hasFeature(Feature.SLICED_DATA_FETCH);
                 this.bucketStrategy = options
                     .getBucketStrategy()
                     .orElseGet(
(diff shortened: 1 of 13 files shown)


heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -210,6 +210,12 @@ private Transform(
             public AsyncFuture<FullQuery> transform(final FindSeries result) throws Exception {
                 final ResultLimits limits;
 
+                if (!slicedFetch) {
+                    return async.resolved(FullQuery.error(namedWatch.end(), QueryError.fromMessage(
+                        "Unable to read metrics data due to mandatory feature 'com.spotify.heroic" +
+                            ".sliced_data_fetch' being disabled.")));
+                }
```
udoprog: I don't think we should error on deprecated features. Either 1) ignore, but that has no deprecation path. 2) Log, but don't spam :). 3) Return it as some kind of non-fatal errors, but not in the `.errors` array, since this is currently considered fatal.
  
Regardless of what we do, I think it is better that happens on the API node (validation). Otherwise each shard would respond with the same thing.

gabrielgerhardsson: Thank you, I agree. 

In my opinion, ideally, a deprecation process should first have a period of time where we only log that a deprecated-and-soon-to-be-removed feature is being used, and then later remove the feature and do something sensible at that time. 
Now, tho, it would be awesome if we can remove this code pretty soon so that the resource tags can be merged 😊  So should we skip straight to remove-code-and-log-when-used (but don't spam)? :) And on the API node, yes 👍

udoprog: Yeah, I'm good with remove and log :+1:

gabrielgerhardsson: Sweet :)

gabrielgerhardsson: I pushed a second commit with a suggestion. Please review.

PR #359, opened 2018-02-06
[metric-bigtable] Dont submit mutateRow request with empty mutations

nezdolik:


metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableBackend.java
```diff
@@ -345,9 +368,12 @@ public Statistics getStatistics() {
 
         for (final Map.Entry<RowKey, Mutations.Builder> e : building.entrySet()) {
             final ByteString rowKeyBytes = rowKeySerializer.serializeFull(e.getKey());
-            writes.add(client
-                .mutateRow(table, rowKeyBytes, e.getValue().build())
-                .directTransform(result -> timer.end()));
+            final Mutations.Builder builder = e.getValue();
+            if (builder.size() > 0) {
```
udoprog: FWIW: This happens because line 354 creates an empty builder when it swaps out the batch. I think I'd prefer killing this section of code in favor of conditionally moving builder into the `saved` list to avoid this duplicate logic.

Does that make sense?

nezdolik:
applied review comment

metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/api/FakeBigtableConnection.java
```diff
@@ -185,7 +182,7 @@ public ColumnFamily createColumnFamily(
                         case SET_CELL:
                             final Mutation.SetCell setCell = mutation.getSetCell();
                             final ColumnFamily columnFamily =
-                                columnFamilies.get(Pair.of(table, setCell.getFamilyName()));
+                                table.getColumnFamily(setCell.getFamilyName()).get();
```
udoprog: use `Optional#orElseThrow()`, otherwise we get a non-descriptive error message when the optional is empty.

PR #360, opened 2018-02-07
Bt client 1.0.0

nezdolik:


metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/api/BigtableMutatorImpl.java
```diff
@@ -166,7 +156,13 @@ public void onFailure(Throwable t) {
 
     private void flush() {
         synchronized (lock) {
-            tableToBulkMutation.values().stream().forEach(BulkMutation::flush);
+            tableToBulkMutation.values().stream().forEach(mutation -> {
+                try {
+                    mutation.flush();
```
udoprog: This didn't use to block, so the `flush` method it's part of didn't block either. We should not change that, since it might eat up threads that are used in other places.

It also blocks under the `BigtableMutatorImpl#lock` Lock, which is used to queue up new mutations and is in the direct write path.

To work around this, we can either use a nonblocking API (if available), or spin up another thread pool with queue to deal with this here.

nezdolik:
Flushing already runs on dedicated executor, added dedicated lock for flush operation.

nezdolik:
Changes (com.spotify.heroic.metric.bigtable.api.BigtableMutatorImpl):
1. Don't se deprecated method when creating BulkMutation object
2. In BigtableMutatorImpl#flush submit bulkMutate requests in async manner per table
3. Don't synchronize when doing flush, as this is already done per BulkMutation object in bt client lib

metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/api/BigtableMutatorImpl.java
```diff
@@ -165,8 +160,36 @@ public void onFailure(Throwable t) {
...
+        final List<AsyncFuture<Void>> futures = tableToBulkMutation
+            .values()
+            .stream()
+            .map(mutation -> async.call((Callable<Void>) () -> {
+                mutation.flush();
+                return null;
+            }))
+            .collect(Collectors.toList());
+
+        async.collect(futures, new StreamCollector<Void, Object>() {
```
udoprog: This is the same as `async.collectAndDiscard(futures)`.

nezdolik:
@udoprog we still need our scheduler to have control over flushing and primarily when performing shutdown. Client lib does not expose methods for proper shutting down and stopping of autoflushing. 
By default autoflushing in lib is enabled if <code>autoflushMs</code> parameter is set in <code>BulkOptions</code>. If we remove executor and not set up this option, no flushing will be performed.

udoprog:
> we still need our scheduler to have control over flushing and primarily when performing shutdown. Client lib does not expose methods for proper shutting down and stopping of autoflushing.

The [`flush`] method performs flushing, and blocks until there are no more pending flushes. This has the side effect of causing the internal executor to be empty.
At this stage, shutdown and awaitTermination is not necessary as long as we don't enqueue more writes. They primarily cause RejectedExecutionException to be raised if the executor is used post shutdown for some reason. Now, we could ask bigtable devs to provide us with a shutdown hook as well, but it's not a blocker IMO.

[`flush`]: https://github.com/GoogleCloudPlatform/cloud-bigtable-client/blob/master/bigtable-client-core-parent/bigtable-client-core/src/main/java/com/google/cloud/bigtable/grpc/async/BulkMutation.java#L442

> By default autoflushing in lib is enabled if autoflushMs parameter is set in BulkOptions. If we remove executor and not set up this option, no flushing will be performed.

If we don't have auto-flushing configured, bulk mutation is not used as per [this line]. This means no BulkMutation instances will be created, so their code is inert. Writes are no longer batched.

[this line]: https://github.com/spotify/heroic/blob/master/metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/api/BigtableMutatorImpl.java#L74

nezdolik:
>If we don't have auto-flushing configured, bulk mutation is not used as per this line. This means no BulkMutation instances will be created, so their code is inert. Writes are no longer batched.

@udoprog bulk mutation is not used in case we have property <code>disableBulkMutations</code> provided in config with 'true' value (which we don't). Autoflushing is not related to this propetry, but to <code>autoflushMs</code> .

>We could ask bigtable devs to provide us with a shutdown hook as well, but it's not a blocker IMO.

I considered this an issue, but if you are fine with non clean shutdown, then would proceed with setting up  <code>autoflushMs</code> property and enabling autoflushing.

udoprog:
> Autoflushing is not related to this propetry

`disableBulkMutations` needs to be entangled with the options we send into the BulkMutator. It doesn't make sense to not have scheduled flushing, and for it to be enabled.

nezdolik:
Switched to autoflushing in client lib. 

*Note: integration test for sending mutation request with size of 4 * MAX_BATCH_SIZE hangs, reducing size to  3 * MAX_BATCH_SIZE makes test to pass. This is not reproduced when performing scheduled flushing in our code. 

Looks like multiple grpc daemon threads are blocked on OS poll call (from thread dump):

<code>
"grpc-default-worker-ELG-2-8" #26 daemon prio=5 os_prio=31 tid=0x00007f80d3bdd800 nid=0x6803 runnable [0x0000700003cf2000]
   java.lang.Thread.State: RUNNABLE
	at sun.nio.ch.KQueueArrayWrapper.kevent0(Native Method)
	at sun.nio.ch.KQueueArrayWrapper.poll(KQueueArrayWrapper.java:198)
	at sun.nio.ch.KQueueSelectorImpl.doSelect(KQueueSelectorImpl.java:117)
	at sun.nio.ch.SelectorImpl.lockAndDoSelect(SelectorImpl.java:86)
	- locked <0x000000077594e820> (a com.spotify.heroic.bigtable.io.netty.channel.nio.SelectedSelectionKeySet)
	- locked <0x000000077594f888> (a java.util.Collections$UnmodifiableSet)
	- locked <0x000000077594e770> (a sun.nio.ch.KQueueSelectorImpl)
	at sun.nio.ch.SelectorImpl.select(SelectorImpl.java:97)
	at com.spotify.heroic.bigtable.io.netty.channel.nio.SelectedSelectionKeySetSelector.select(SelectedSelectionKeySetSelector.java:62)
	at com.spotify.heroic.bigtable.io.netty.channel.nio.NioEventLoop.select(NioEventLoop.java:753)
	at com.spotify.heroic.bigtable.io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:409)
	at com.spotify.heroic.bigtable.io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:858)
	at com.spotify.heroic.bigtable.io.netty.util.concurrent.DefaultThreadFactory$DefaultRunnableDecorator.run(DefaultThreadFactory.java:138)
	at java.lang.Thread.run(Thread.java:748)
</code>

gabrielgerhardsson:
@nezdolik Hmm interesting. The grpc threads being blocked like that, I'd say it's likely that they are waiting for data from the tcp/ip socket. I.e., that would mean that any issue is on the BT backend side.

nezdolik:
WP: Excluding this CP part <code>io.netty.util.internal.*</code> from relocation creates shared context for netty-commons in grpc and repackaged-bigtable modules. That's why we need to use consistent versions of netty and grpc across project. This solution still fails on grpc tests with DEADLINE_EXCEEDED, which is a misleading message.

dmichel1:
closing this as the BT client was updated to 1.4.0 [here](https://github.com/spotify/heroic/pull/374/commits/c7886ef144bb249f957073c32eb5a6833b5b0df0)

One thing I'm uncertain of is if we can remove the [scheduled executor](https://github.com/spotify/heroic/blob/master/metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/api/BigtableMutatorImpl.java#L75) for flushing since the client does that now (?)

PR #365, opened 2018-05-24
Calculate delta per second

nabam:
This should make delta values independent from resolution

### DIFF
--- a/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/DeltaInstance.java
+++ b/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/DeltaInstance.java
@@ -94,7 +94,9 @@ public class DeltaInstance implements AggregationInstance {
         while (it.hasNext()) {
             final Point current = it.next();
             double diff = current.getValue() - previous.getValue();
-            result.add(new Point(current.getTimestamp(), diff));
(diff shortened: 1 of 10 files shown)


nezdolik:
Change needs to be backwards compatible, not to impact existing users.

dmichel1:
@nezdolik can we get this merged since the feature has already been released internally?

dmichel1:
@nabam @nezdolik  was the `NotNegative` aggregation added to get around the issue that `Delta` and `DeltaPerSecond` can produce negative values?

I'm looking at adding two more aggregations that are similar to prometheus, `rate()` and `irate()` that work with counters and can handle them being reset without producing negative values.

https://github.com/prometheus/prometheus/blob/2bd510a63e48ac6bf4971d62199bdb1045c93f1a/docs/querying/functions.md#rate

https://github.com/prometheus/prometheus/blob/c6399296dcc059bbc496f40e31aa2c7b7a341951/promql/functions.go#L178

nabam:
@dmichel1 `Average(NotNegative(DeltaPerSecond()))` is essentially a prometheus `rate`. NotNegative is used to drop data points when overflow or restart happens. Without NotNegative it can be used on gauges to get derivatives.

dmichel1:
as discussed - using `NotNegative` will result in a loss of precision when the counter resets. Seems like creating rate/irate is the way forward.

PR #366, opened 2018-06-07
Update ElasticSearch Templates to reflect deprecation of "string"

jcabmora:
As of ElasticSearch 5.0, elastic has expressed the intention to deprecate the `string` type in favor of two new data types `text` and `keyword`:  https://www.elastic.co/blog/strings-are-dead-long-live-strings

This PR reflects that change by replacing `not_analyzed` strings as `keyword` type, and `analyzed` strings with the `text` type.  It also simplifies the templates a little bit by disabling `_all` at the parent type level since all of the fields had the `include_in_all` parameter set to false.

### DIFF
--- a/metadata/elasticsearch/src/main/resources/com.spotify.heroic.metadata.elasticsearch/kv/metadata.json
+++ b/metadata/elasticsearch/src/main/resources/com.spotify.heroic.metadata.elasticsearch/kv/metadata.json
@@ -1,23 +1,20 @@
 {
   "metadata": {
+    "_all": {
+        "enabled": false
+    },
     "properties": {
       "key": {
-        "index": "not_analyzed",
-        "type": "string",
-        "doc_values": true,
-        "include_in_all": false
+        "type": "keyword",
+        "doc_values": true
       },
       "tags": {
-        "type": "string",
-        "index": "not_analyzed",
-        "doc_values": true,
-        "include_in_all": false
+        "type": "keyword",
+        "doc_values": true
       },
       "tag_keys": {
-        "type": "string",
-        "index": "not_analyzed",
-        "doc_values": true,
-        "include_in_all": false
+        "type": "keyword",
+        "doc_values": true
       }
     }
   }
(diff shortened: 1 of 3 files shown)


udoprog:
@jcabmora This looks great!

I am a little bit concerned this will prevent folks from configuring pre `keyword` and `text` clusters.

Could we either detect which schema a cluster supports and adapt accordingly, or make it a non-default configuration option for now until we officially bump which ES versions we are compatible with?

Thanks!

jcabmora:
@udoprog The problem that I see is that I think the current code base won't work with an ES 2.X cluster. If that is true, then I believe there is no point to try to maintain backwards compatibility for ES templates.

I am not particularly knowledgeable in ElasticSearch, so I could be wrong.  However, I did run a test using an ES 2.2.1 ES cluster and a heroic jar built with the current HEAD, and when I ran the `configure` HeroicShell task there were multiple exceptions reported and the templates were not created.  Here's an extract of the logs:
```
21:27:11.654 [elasticsearch[_client_][transport_client_boss][T#1]] WARN  org.elasticsearch.transport.netty4.Netty4Transport - exception caught on transport layer [[id: 0x2fade185, L:/172.16.0.45:44622 - R:elasticsearchclient/172.16.0.48:9300]], closing connection
java.io.EOFException: tried to read: 36 bytes but only 3 remaining
```
The ES docs are vague about this (https://www.elastic.co/guide/en/elasticsearch/client/java-api/5.3/client.html), but it seems that the client must match the version of the server.

udoprog:
@jcabmora good point! we should update the documentation that we no longer support ES 2. Thanks!

jcabmora:
Is this PR going to be merged? We want to update our internal version, and it will be nice to have this so our code is easier to rebase.

dmichel1:
@jcabmora sorry for the delay - if you rebase against master I'll merge this right away. 

I'll look over #371 next. It will also need a rebase once this is merged. Thanks!

jcabmora:
@dmichel1 Thanks a lot ! I have rebased my PR branch so it is at the same HEAD as master. Thanks!!

PR #379, opened 2018-09-06
Add consumer metric reporting to PubSub.

hexedpackets:


consumer/pubsub/src/main/java/com/spotify/heroic/consumer/pubsub/Receiver.java
```diff
@@ -43,14 +57,24 @@ public void receiveMessage(PubsubMessage message, AckReplyConsumer replyConsumer
...
+            consumed.increment();
+            replyConsumer.ack();
+        } catch (ConsumerSchemaValidationException e) {
+            reporter.reportConsumerSchemaError();
             log.error("ID:{} - {}", messageId, e.getMessage(), e);
-        } finally {
+
+            // The message will never be processable, ack it to make it go away
             replyConsumer.ack();
+        } catch (Exception e) {
```
dmichel1: for my understanding - removing the ack from the finally block will put the message back onto the queue to be delivered at another time, right?

PR #382, opened 2018-09-11
Fix dependancy conflict for histograms using MinMaxSlidingTimeReservoir.

hexedpackets:


dmichel1:
@hexedpackets can we create an issue to see if we still need this MinMaxSlidingTimeReservoir? 

\cc @gabrielgerhardsson

PR #383, opened 2018-09-12
[heroic-elasticsearch-utils] add metrics to DisabledRateLimitedCache

dmichel1:
The `DisabledRateLimitedCache` was missing metrics around how often it was being hit.

@hexedpackets @jsferrei

### DIFF
--- a/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/DisabledRateLimitedCache.java
+++ b/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/DisabledRateLimitedCache.java
@@ -24,13 +24,22 @@ package com.spotify.heroic.elasticsearch;
 import java.util.concurrent.ConcurrentMap;
 import lombok.RequiredArgsConstructor;
 
+/**
+ * A cache to reduce writes to Elasticsearch.
+ *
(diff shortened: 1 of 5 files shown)


heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/DisabledRateLimitedCache.java
```diff
@@ -24,13 +24,22 @@
...
 @RequiredArgsConstructor
 public class DisabledRateLimitedCache<K> implements RateLimitedCache<K> {
     private final ConcurrentMap<K, Boolean> cache;
 
     @Override
     public boolean acquire(K key, final Runnable cacheHit) {
-        return cache.putIfAbsent(key, true) == null;
+        if (cache.putIfAbsent(key, true) == null) {
+            cacheHit.run();
+            return false;
```
hexedpackets: The return values here are inversed from what they were before, is that intentional?

dmichel1: no, its not 👎 - that would explain the graphs from yesterday. I added a unit test and fixed the conditional.

PR #384, opened 2018-09-13
[docs] es mappings + bigtable timestamps

dmichel1:
* add info about needing ES settings/mapping.
* add note about Bigtable mutating rows and how that prevents duplicate metrics that have the same timestamp.

@hexedpackets @jsferrei @lmuhlha @sjoeboo


lmuhlha:
+1

PR #387, opened 2018-09-14
[heroic-elasticsearch-utils] add DistributedRateLimitedCache

dmichel1:
A distributed cache that does not allow to be called more than a specific rate.

Instead of having a cache that is local to each consumer this allows the cache to be distributed to all the instances. This became necessary when migrating from Kafka -> Google PubSub as a consumer with PubSub will not have a stable partition of metrics.

Reads from the distributed cache are stored locally to reduce the load on the distributed cache cluster.

There is a separate commit that updates folsom to the latest stable version.

@hexedpackets @jsferrei @sjoeboo

### DIFF
--- a/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/DistributedRateLimitedCache.java
+++ b/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/DistributedRateLimitedCache.java
@@ -1,0 +1,117 @@
+/*
+ * Copyright (c) 2018 Spotify AB.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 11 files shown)


heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/DistributedRateLimitedCache.java
```diff
@@ -0,0 +1,116 @@
...
+      try {
+        if (memcachedClient.get(
+          cacheKey).toCompletableFuture().get(100, TimeUnit.MILLISECONDS) != null) {
+          span.addAnnotation("Found key in memcached");
+          cache.putIfAbsent(key, true);
+          cacheHit.run();
+          return false;
+        }
+      } catch (TimeoutException | InterruptedException | ExecutionException e) {
+        span.addAnnotation("Failed to get key from memcached");
```
hexedpackets: Would also be nice to set the span status here to make it more visible that it failed.

dmichel1: 👍 - added  `span.setStatus(Status.INTERNAL);`

PR #388, opened 2018-09-18
[pubsub] consume more messages

dmichel1:
* Change the default number of outstanding messages to 20k.
* Set parallel pull count to the number of configured threads.

throughput improvements of a consumer compared to its peers. 
throughput improvement

@sjoeboo @hexedpackets


sjoeboo:
+1

PR #389, opened 2018-09-19
Add HTTP tracing in jersey for incoming requests.

hexedpackets:
Based on the existing Jersey OpenTracing implementation: https://github.com/eclipse-ee4j/jersey/tree/master/incubator/open-tracing/src/main/java/org/glassfish/jersey/opentracing

### DIFF
--- a/heroic-core/pom.xml
+++ b/heroic-core/pom.xml
@@ -61,6 +61,10 @@
       <artifactId>jetty-rewrite</artifactId>
     </dependency>
     <dependency>
+    <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-util</artifactId>
+    </dependency>
+    <dependency>
       <groupId>org.glassfish.jersey.containers</groupId>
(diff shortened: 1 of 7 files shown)


hexedpackets:
I'm going to move this to a separate library, closing for now.

hexedpackets:
@dmichel1 I question how much value unit tests would add here. Since the new classes are strictly doing side effects based on fired events, unit tests would just look like copies of the OpenCensus API and would be tightly coupled to the implementation.

I think it would be great if we could do some kind of integration test here instead - eg verify that when we send a GET to `/status`, a trace is created containing at least two spans (jetty server request and the heroic resource itself). I think it would be a decent chunk of work to setup the initial suite for that, but its worth doing at some point.

dmichel1:
@hexedpackets I think it being tied to the implementation of the Jersey part of the tracing is okay. there are a lot of attributes being set right now based on various conditions. Once we start extracting header information there will be even more areas to test to make sure the right thing is happening.

PR #395, opened 2018-11-09
Bump folsom to 1.2.1 as it fixes a connection leak

jsferrei:
@dmichel1 I'm going to run some quick tests on this first but https://github.com/spotify/folsom/issues/103 has been fixed  in 1.2.1


dmichel1:
LGTM - 

re: memcached metrics there is also a way to enable the semantic metrics that are part of the folsom client, https://github.com/spotify/folsom/blob/master/folsom-semantic-metrics/src/main/java/com/spotify/folsom/client/SemanticFolsomMetrics.java.

Your approach allows for finer grained metrics since its per consumer-type.

PR #397, opened 2018-11-26
add protobuf schema; deprecate host field

dmichel1:
Two commits.

1. Add protobuf schema for ingesting metrics.
2. deprecate host field in json schema since host is now just another tag in heroic. 

@jsferrei @lmuhlha @sjoeboo

### DIFF
--- a/heroic-core/src/test/java/com/spotify/heroic/consumer/schemas/Spotify100ProtoTest.java
+++ b/heroic-core/src/test/java/com/spotify/heroic/consumer/schemas/Spotify100ProtoTest.java
@@ -1,0 +1,122 @@
+/*
+ * Copyright (c) 2015 Spotify AB.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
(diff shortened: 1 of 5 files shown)


heroic-core/src/test/java/com/spotify/heroic/consumer/schemas/Spotify100ProtoTest.java
```diff
@@ -0,0 +1,98 @@
...
+  @Before
+  public void setup() {
+    when(clock.currentTimeMillis()).thenReturn(1542830485000L);
+    when(ingestion.write(any(Ingestion.Request.class))).thenReturn(resolved);
+    consumer = new Spotify100Proto.Consumer(clock, ingestion, reporter);
+  }
+
+  @Test
+  public void testConsumeMetric() throws Exception {
+    final Metric metric = Metric.newBuilder()
```
hexedpackets: Can we hard code the bytes here instead? Using the builder doesn't seem like a great test - its just proving that the proto definition can serialize and deserialize itself. I'd rather have a test that fails if we make an incompatible change to the definition or the consumer.

dmichel1: Sure - I can add an additional test with the raw-bytes.

dmichel1:
@hexedpackets any other feedback about this before merging?

PR #398, opened 2018-11-27
bugfix for docker build (per

sjoeboo:
https://stackoverflow.com/questions/53010200/maven-surefire-could-not-find-forkedbooter-class/53085816#53085816)


dmichel1:
Can you fix the title of the PR? looks like it got truncated.

+1 after that.

PR #399, opened 2018-11-28
add tag to trigger for docker hub

sjoeboo:
This change allows for a few things: 

* Travis will get triggered on tags matching master-* as well as master.
* Travis will trigger a Docker image build on Docker Hub *only* from tags (releases)
* Travis will send the tag to Docker Hub so the resulting image is tagged the same.


dmichel1:
+1 after squashing down the 2 commits and adding a commit message with what you are trying to achieve.

sjoeboo:
@dmichel1 updated

PR #401, opened 2018-12-19
[pubsub] add additional logging when fatal error occurs

dmichel1:
@lmuhlha @sjoeboo @jsferrei @hexedpackets


consumer/pubsub/src/main/java/com/spotify/heroic/consumer/pubsub/Connection.java
```diff
@@ -130,6 +132,16 @@ Connection start() {
             .setCredentialsProvider(credentialsProvider)
             .build();
 
+        subscriber.addListener(
+          new Subscriber.Listener() {
+              @Override
+              public void failed(Subscriber.State from, Throwable failure) {
+                  // Called when the Subscriber encountered a fatal error and is shutting down
+                  log.error(
```
dmichel1: I'm open to exiting the process since we CANNOT recover from this but this will also mask 1 of the issues we've been having with Google.

dmichel1: @hexedpackets ptal - added 1 line :)

hexedpackets: :+1:

PR #403, opened 2019-01-03
[core] modify proto schema to accept compressed batches

dmichel1:
When rolling this out we may need to tweak both the publisher (ffwd) and consumer (heroic) batch settings. With the consumer we may want to reduce the maxOutstandingElement count since a single pubsub message can now contain up to 10,000 metrics with the default ffwd batch settings.

This goes along with https://github.com/spotify/ffwd/pull/99.

@hexedpackets

### DIFF
--- a/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/DistributedRateLimitedCache.java
+++ b/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/DistributedRateLimitedCache.java
@@ -94,7 +94,7 @@ public class DistributedRateLimitedCache<K> implements RateLimitedCache<K> {
       } catch (TimeoutException e) {
         span.setStatus(Status.INTERNAL.withDescription(e.getMessage()));
         memcachedReporter.reportMemcachedTimeout();
-        log.error("Failed to get key from memecached");
+        log.debug("Failed to get key from memecached");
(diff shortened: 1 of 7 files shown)


heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/DistributedRateLimitedCache.java
```diff
@@ -94,7 +94,7 @@ public boolean acquire(K key, final Runnable cacheHit) {
       } catch (TimeoutException e) {
         span.setStatus(Status.INTERNAL.withDescription(e.getMessage()));
         memcachedReporter.reportMemcachedTimeout();
-        log.error("Failed to get key from memecached");
+        log.debug("Failed to get key from memecached");
```
hexedpackets: How come you made this a debug statement? It looks like a real error trying to communicate with memcached.

dmichel1: to reduce the spam in the logs - found this less important now that we have a metric reporting the same thing. 

iirc this is mostly happening because we don't use the async framework properly in this module, as this error will be raised when it hits `.toCompletableFuture().get(100, TimeUnit.MILLISECONDS)`

PR #407, opened 2019-02-05
Add tracing to heroic-api

lmuhlha:


heroic-core/pom.xml
```diff
@@ -49,8 +49,16 @@
     <dependency>
       <groupId>com.google.guava</groupId>
       <artifactId>guava</artifactId>
+      <version>27.0-jre</version>
```
dmichel1: all the versions should be controlled in the parents pom dependendency management section

heroic-core/pom.xml
```diff
@@ -49,8 +49,16 @@
...
     </dependency>
 
+    <dependency>
+      <groupId>com.google.cloud</groupId>
+      <artifactId>google-cloud-core</artifactId>
+      <version>1.53.0</version>
```
dmichel1: should move this version there as well.

heroic-core/pom.xml
dmichel1: https://github.com/spotify/heroic/blob/master/pom.xml#L545

heroic-core/pom.xml
```diff
@@ -149,9 +157,87 @@
...
+      <version>0.0.2</version>
+    </dependency>
+
+    <dependency>
+      <groupId>io.opencensus</groupId>
+      <artifactId>opencensus-exporter-trace-jaeger</artifactId>
+    </dependency>
+
+    <!-- opencensus exporter dependencies -->
+    <dependency>
```
dmichel1: this is already declared above

heroic-core/pom.xml
```diff
@@ -149,9 +157,87 @@
...
+      <groupId>com.google.protobuf</groupId>
+      <artifactId>protobuf-java</artifactId>
+      <version>3.6.1</version>
+    </dependency>
+
+    <dependency>
+      <groupId>com.google.cloud</groupId>
```
dmichel1: same with this one.

heroic-core/src/main/java/com/spotify/heroic/HeroicCore.java
```diff
@@ -137,6 +140,11 @@
     static final boolean DEFAULT_DISABLE_BACKENDS = false;
     static final boolean DEFAULT_SETUP_SHELL_SERVER = true;
 
+    private static final String DEFAULT_COLLECTOR_HOST = "lightstep-satellite.services.heroic";
```
dmichel1: since collector host and token are both required there's no need to have default options.

heroic-core/src/main/java/com/spotify/heroic/HeroicCore.java
```diff
@@ -260,21 +268,59 @@ public HeroicCoreInstance newInstance() throws Exception {
...
     @SuppressWarnings("unchecked")
     private void enableTracing(Map<String, Object> config) throws IOException {
-        final Object stackdriverConfig = config.get("stackdriver");
-        if (stackdriverConfig != null) {
-            final Map<String, String> configMap = (Map) stackdriverConfig;
-            final String projectId = configMap.get("project");
+        final Object lightstepConfig = config.get("lightstep");
+        if (lightstepConfig != null) {
+            final Map<String, String> configMap = (Map) lightstepConfig;
+            final String collectorHost = configMap.getOrDefault("collectorHost",
```
dmichel1: you can check now to see if collectorHost and accessToken are null and throw an exception that the required parameters are missing.

heroic-core/src/main/java/com/spotify/heroic/HeroicCore.java
```diff
@@ -260,21 +268,59 @@ public HeroicCoreInstance newInstance() throws Exception {
...
+                DEFAULT_COLLECTOR_HOST);
+            final String accessToken = configMap.getOrDefault("accessToken", DEFAULT_LIGHTSTEP_TOKEN);
+            final Integer reportingIntervalMs = (Integer) config.getOrDefault("reportingIntervalMs",
+                DEFAULT_LIGHTSTEP_REPORTING_MS);
+            final Integer maxBufferedSpans = (Integer) config.getOrDefault("maxBufferedSpans",
+                DEFAULT_LIGHTSTEP_MAX_SPANS);
+
+
+            if (collectorHost == null) {
```
dmichel1: Which you started to do here already but are missing for accessToken.

heroic-core/src/main/java/com/spotify/heroic/HeroicCore.java
```diff
@@ -260,21 +268,59 @@ public HeroicCoreInstance newInstance() throws Exception {
...
+
+        }
 
-            if (projectId == null) {
-                throw new IllegalArgumentException("Configuration for stackdriver is incomplete");
+        final Object signalfxConfig = config.get("signalfx");
+        if (signalfxConfig != null) {
+            final Map<String, String> configMap = (Map) signalfxConfig;
+            final String region = configMap.get("region");
+            final String defaultEndpoint = configMap.getOrDefault(format("http://signalfxgateway.services"
```
dmichel1: this forces anyone using heroic to have a similar naming convention for the signalfx gateway. it would be better to follow a similar pattern to lightstep and just have `collectorHost` or something similar

dmichel1: also since this is a required option there should be no default and you can throw an exception

pom.xml
```diff
@@ -530,6 +530,54 @@
...
+      <dependency>
+        <groupId>com.lightstep.opencensus</groupId>
+        <artifactId>lightstep-opencensus-exporter</artifactId>
+        <version>0.0.2</version>
+      </dependency>
+
+      <dependency>
+        <groupId>io.opencensus</groupId>
+        <artifactId>opencensus-impl</artifactId>
+        <version>0.19.0</version>
```
dmichel1: this should use the ${opencensus.version} property

heroic-core/pom.xml
```diff
@@ -149,9 +157,87 @@
...
       <artifactId>protobuf-java</artifactId>
-      <version>3.5.1</version>
+      <version>3.6.1</version>
+    </dependency>
+
+    <!-- opencensus dependencies -->
+    <dependency>
+      <groupId>io.opencensus</groupId>
+      <artifactId>opencensus-api</artifactId>
+      <version>${opencensus.version}</version>
```
dmichel1: you can get rid of all these versions since they are declared already in the parent pom.

heroic-core/src/main/java/com/spotify/heroic/http/HttpServer.java
```diff
@@ -89,7 +89,8 @@ public HttpServer(
         @Named("enableCors") final boolean enableCors,
         @Named("corsAllowOrigin") final Optional<String> corsAllowOrigin,
         final List<JettyServerConnector> connectors,
-        @Named("stopping") final Supplier<Boolean> stopping
+        @Named("stopping") final Supplier<Boolean> stopping,
+        @Named("tracing") final Optional<String> tracing
```
hexedpackets: These options come from dagger, and I don't see any provider to inject `tracing`. It also isn't used for anything so you can get rid of it.

heroic-core/src/main/java/com/spotify/heroic/http/tracing/OpenCensusApplicationEventListener.java
```diff
@@ -278,6 +279,7 @@ private void log(OpenCensusFeature.Verbosity level, String s) {
         private void logError(final Throwable t) {
             requestSpan.setStatus(Status.INTERNAL);
             requestSpan.putAttribute("event", AttributeValue.stringAttributeValue("error"));
+            requestSpan.putAttributes(environmentMetadata.toAttributes());
```
hexedpackets: Having the metadata here will only attach it when a request throws an error. You should move it to the span start so its on all of them - probably the request span, not the resource span, since the request span will always be created.

lmuhlha:
@dmichel1 @hexedpackets made the changes as requested. :)

hexedpackets:
@lmuhlha I think you might have missed my comment about parsing the incoming HTTP headers. The rest looks good :+1:

lmuhlha:
@hexedpackets yeah sorry i was just looking at the file changes when going through this. will work on that as well.

lmuhlha:
@hexedpackets I added the part to parse the headers, not entirely sure how to account for that afterwards.

hexedpackets:
@lmuhlha You're adding the context to the headers right before the Java method takes over. That isn't going to do anything, the span was already started at that point and you would only want to inject it into headers for a client wrapper. We need to do the inverse - read the context from the headers and start the span as a child of what gets parsed, if there's anything there.

lmuhlha:
@hexedpackets Minus getting the extract working properly, I think this is good now

lmuhlha:
!rebuild

PR #408, opened 2019-02-06
[aggregation] add rate and ratePerSecond aggregations

dmichel1:
These aggregations are to be used with counters. They differ from delta aggregations because they do not produce negative values and can handle counter resets/

I will update the documentation on how to use these.

@hexedpackets @jsferrei @lmuhlha


aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/RateInstance.java
```diff
@@ -0,0 +1,176 @@
...
+        final Iterator<Point> it = points.iterator();
+        final ArrayList<Point> result = new ArrayList<>();
+
+        if (!it.hasNext()) {
+            return Collections.emptyList();
+        }
+
+        Point previous = it.next();
+
+        while (it.hasNext()) {
```
hexedpackets: If there's only one point, does it make sense to return an empty array, or should we return an array with that point to indicate the rate as (point-0)?

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/RateInstance.java
```diff
@@ -0,0 +1,176 @@
...
+            final Point current = it.next();
+
+            // Counter reset.
+            if (current.getValue() < previous.getValue()) {
```
hexedpackets: This doesn't necessarily mean the counter was reset. Counters can be decremented, and I think this won't calculate the rate correctly in the case.

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/RatePerSecondInstance.java
```diff
@@ -0,0 +1,177 @@
...
+/**
+ * A rate per second implementation inspired by the prometheus irate() function.
+ *
+ * It is intended to be be used with counters that are increasing to calculate the rate of change
+ * per second difference between two points. It handles the counter being reset.
+ *
+ * @author dmichel
+ */
+@Data
+public class RatePerSecondInstance implements AggregationInstance {
```
hexedpackets: Do we need a special case for this? I think having both will be confusing for people trying to decide which one to use. What are the different use cases you're picturing for someone to want rate vs rate per second?

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/RateInstance.java
dmichel1: you are right i'll fix this comment...

in the docs for this aggregation I will explain this should only be used with monotonically increasing counters. this is the same as prom https://github.com/prometheus/prometheus/blob/2bd510a63e48ac6bf4971d62199bdb1045c93f1a/docs/querying/functions.md#rate

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/RatePerSecondInstance.java
dmichel1: I think we do - this is similar to the delta and deltaPerSecond aggregations we have already.

for SLO calculates over long time periods I envision using just the rate aggregation to get a total sum.

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/RateInstance.java
dmichel1: hmm, i don't think we can return anything but an empty array because returning a new `Point()` requires a timestamp which we don't have.

PR #409, opened 2019-02-07
Update semantic-metrics and replace meters with counters

hexedpackets:
We've found meters to be difficult for people to understand, in particular the difficulties around aggregating the results tends to be hard to explain. Using counters and calculating aggregated rates from the raw counts is more aligned with how people intuitively understand the metrics.

This will be a breaking change as far as our graphs and monitoring go. I'd like to get it merged in but wait to release it until we're ready to cut a major version.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,7 @@
     <jackson.version>2.8.11</jackson.version>
     <jetty.version>9.4.12.v20180830</jetty.version>
     <lucene.version>5.5.0</lucene.version>
-    <semantic-metrics.version>0.2.3</semantic-metrics.version>
+    <semantic-metrics.version>1.0.2</semantic-metrics.version>
     <netty.version>4.1.31.Final</netty.version>
     <tiny-async.version>1.11.0</tiny-async.version>
     <tiny-serializer.version>1.4.5</tiny-serializer.version>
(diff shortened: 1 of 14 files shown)


pom.xml
```diff
@@ -804,6 +804,18 @@
         <scope>compile</scope>
       </dependency>
 
+      <!-- metrics -->
+      <dependency>
+        <groupId>com.spotify.metrics</groupId>
```
dmichel1: fyi @stas added a BOM to ffwd so we don't have to have multiple dependencies. just need to release it.

https://github.com/spotify/semantic-metrics/blob/master/semantic-metrics-bom/pom.xml

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/HistogramBuilder.java
```diff
@@ -25,13 +25,15 @@
 import com.codahale.metrics.ExponentiallyDecayingReservoir;
 import com.codahale.metrics.Histogram;
 import com.codahale.metrics.Metric;
+import com.spotify.metrics.core.HistogramWithTtl;
```
dmichel1: looks like you aren't using `HistogramWithTtl` yet?

PR #410, opened 2019-02-12
Use protobuf for shell serialization.

hexedpackets:


heroic-component/src/main/java/com/spotify/heroic/shell/ShellIO.java
```diff
@@ -43,19 +42,17 @@
      * Open a file for reading.
      *
      * @param path Path of file to open.
-     * @param options Options when opening file.
      * @return An InputStream associated with the open file.
      */
-    InputStream newInputStream(Path path, StandardOpenOption... options) throws IOException;
```
dmichel1: What were these options and how/do they still work?

heroic-core/src/main/java/com/spotify/heroic/CoreShellTasks.java
```diff
@@ -133,15 +136,15 @@ public PrintWriter out() {
             }
 
             @Override
-            public OutputStream newOutputStream(Path path, StandardOpenOption... options)
```
dmichel1: ditto here a about the options

heroic-component/src/main/java/com/spotify/heroic/shell/ShellIO.java
hexedpackets: The `StandardOpenOption` enumerates the POSIX params for opening a file - read, write, append, etc.

AFAICT the only thing that was ever passed in was `[READ]` so I removed it as an option.

PR #412, opened 2019-02-14
[elasticsearch-utils] enable dynamic lookup of nodes

dmichel1:
Currently, heroic is deployed internally with the list of seeds being all the data nodes.

During an incident when capacity was added to an already "red" cluster, puppet modified the heroic config with the new data nodes, and restarted the api. The API was never able to be successfully started bringing down everything.

This change will allow us to have the master nodes that rarely change become the seeds and the data nodes "sniffed" out. Sniffing is disabled by default and will have to be explicitly enabled. If enabled new nodes will be polled for by default every 30 seconds.

https://www.elastic.co/guide/en/elasticsearch/client/java-api/current/transport-client.html

@lmuhlha @hexedpackets @sjoeboo

### DIFF
--- a/example/heroic.example.yml
+++ b/example/heroic.example.yml
@@ -91,9 +91,10 @@ metadata:
         # Random seed, set if you want consistency.
         # @default Randomly generated.
         seed: 0
+
     ## ElasticSearch-based metadata.
-    # - backendType: kv
-    #   # deprecated but useful as a descriptor
+    # - type: elasticsearch
+    #   backendType: kv
     #   type: elasticsearch
     #   # Backend id, if not specified it will be generated.
     #   # @default null
@@ -105,6 +106,13 @@ metadata:
     #     clusterName: elasticsearch
     #     # ElasticSearch index.
     #     # @default "heroic"
+    #
+    #     # Dynamically sniff new nodes. Defaults false
+    #     sniff: false
+    #
+    #     How often to sniff for new nodes. Defaults 30 seconds
+    #     nodeSamplerInterval: 30s
+    #
     #     index:
     #       # Index mapping to use
     #       # @default "single": only operates on a single index.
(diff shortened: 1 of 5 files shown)


dmichel1:
@hexedpackets I made sniffing disabled by default and made the default sniff interval 30 seconds.

example/heroic.example.yml
```diff
@@ -105,6 +106,13 @@ metadata:
     #     clusterName: elasticsearch
     #     # ElasticSearch index.
     #     # @default "heroic"
+    #
+    #     # Dynamically sniff new nodes. Defaults false
+    #     sniff: false
+    #
+    #     How often to sniff for new nodes. Defaults 30 seconds
+    #     nodeSamplerInterval: 30
```
hexedpackets: The new params should be added to the docs too

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/TransportClientSetup.java
```diff
@@ -36,21 +37,39 @@
     public static final String DEFAULT_CLUSTER_NAME = "elasticsearch";
     public static final List<String> DEFAULT_SEEDS = ImmutableList.of("localhost");
     public static final int DEFAULT_PORT = 9300;
+    public static final boolean DEFAULT_SNIFF = true;
```
hexedpackets: You're defaulting sniff to true here. I'm fine with having true as the default, but the example config lists it as false.

dmichel1: I wanted to leave that as false. fixed it here.

example/heroic.example.yml
dmichel1: Added them to the docs. So nice they are on the same branch!

dmichel1:
@hexedpackets it built successfully on openjdk8, merging.

PR #413, opened 2019-02-14
 Replace lombok JSONified data classes with autovalue.

hexedpackets:
There's still a lot of lombok in the code that will need to get replaced in a future PR for Java 9+. Upgrading lombok broke a lot of data classes, seemingly all the ones that act as JSON data objects. I replaced many of those with AutoValue and removed any unnecessary lombok annotations in the classes. After doing that, I actually un-upgraded lombok since there were still a bunch of things failing and this PR is already quite huge.

I've also removed any tests that were solely checking whether the lombok `@Data` annotation worked. The test helper for that used a bunch of reflection that breaks in Java 9+, and doesn't seem to add value to the test suite. If those classes compile, they would pass the test (at least in Java 8).


hexedpackets:
Ah, that makes sense. I should go back through and update those tests then.

PR #416, opened 2019-02-19
Delombokify constructor annotations.

hexedpackets:
The java.beans annotations are not really necessary, they're a byproduct of running delombok in these classes. They shouldn't have any downside other than adding verbosity, so it was easier to leave them.


dmichel1:
I think I would prefer to remove the beans if they aren't necessary. They clutter the code a bit.

hexedpackets:
> I think I would prefer to remove the beans if they aren't necessary. They clutter the code a bit.

I looked into this and I understand now what they're doing and why they're necessary. The `ConstructorProperties` annotation is generating getters (and setters in theory, but I don't think any of the classes with them are mutable), and creating a mapping for serialization/deserialization. So sort of a combination of jackson's `@JsonCreator` property plus getter generation.

I would hazard a guess that its not needed everywhere that its set, but I think it would be quite a bit of work to go through everything and figure out which annotations are superfluous.

PR #417, opened 2019-02-20
Replace data classes with kotlin

hexedpackets:
Only a couple of classes are converted in this PR, but I wanted to open it now since it adds kotlin to the maven builds and jackson serialization.


heroic-core/src/main/java/com/spotify/heroic/shell/ShellServer.java
```diff
@@ -91,7 +91,7 @@ void doRun(ShellServerState state) {
             final Socket socket;
 
             try {
-                socket = state.serverSocket.accept();
+                socket = state.getServerSocket().accept();
```
dmichel1: what is this about?

hexedpackets: The variable used to be package local but nothing ever tried to set it after initialization so I made it private (which causes both kotlin and lombok to generate getters). I'm fine with making it package local again if you think we should try to keep parity with the old code during this refactor.

Issue #96, opened 2016-10-09
Fix or filter support in MetadataBackendV1


udoprog:
It was discovered after implementing a more comprehensive test suite for metadata backends that or filtering appears to _not work_ in V1 (soon to be deprecated). This should be fixed until V1 is actually deprecated.

See:
https://github.com/spotify/heroic/blob/master/metadata/elasticsearch/src/test/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendV1IT.java#L39

PR #418, opened 2019-02-22
[metadata] delete deprecated v1 backend

dmichel1:
V1 has long been dead and replaced with KV.

@hexedpackets @jsferrei @sjoeboo


dmichel1:
Fixes https://github.com/spotify/heroic/issues/96

PR #420, opened 2019-02-27
Tracing bug fixes

lmuhlha:
- Autovalue import was missing for EnvironmentMetadata
- Handle Text Format Getter null pointer exception
- Update config example
- Tracing code clean up
- Link Query Resource spans
- Lightstep should be configured to load balance correctly
- Update gradle dependencies

### DIFF
--- a/heroic-core/src/main/java/com/spotify/heroic/HeroicCore.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/HeroicCore.java
@@ -354,7 +354,7 @@ public class HeroicCore implements HeroicConfiguration {
         }
 
         final double probability = (double) config.getOrDefault("probability", 0.01);
-        log.info("Setting tracing to sample with a probability of {}", String.valueOf(probability));
+        log.info("Setting tracing to sample with a probability of {}", probability);
         TraceConfig traceConfig = Tracing.getTraceConfig();
         traceConfig.updateActiveTraceParams(
             traceConfig
--- a/heroic-core/src/main/java/com/spotify/heroic/HeroicCore.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/HeroicCore.java
@@ -84,8 +84,6 @@ import eu.toolchain.async.AsyncFramework;
 import eu.toolchain.async.AsyncFuture;
 import eu.toolchain.async.ResolvableFuture;
 import io.opencensus.contrib.zpages.ZPageHandlers;
-import io.opencensus.exporter.trace.jaeger.JaegerTraceExporter;
-import io.opencensus.exporter.trace.zipkin.ZipkinTraceExporter;
 import io.opencensus.trace.Tracing;
 import io.opencensus.trace.config.TraceConfig;
 import io.opencensus.trace.samplers.Samplers;
@@ -139,6 +137,8 @@ public class HeroicCore implements HeroicConfiguration {
 
     private static final int DEFAULT_LIGHTSTEP_REPORTING_MS = 1_000;
     private static final int DEFAULT_LIGHTSTEP_MAX_SPANS = 1_000;
+    private static final boolean DEFAULT_GRPC_ROUNDROBIN = true;
+    private static final boolean DEFAULT_GRPC_RESETCLIENT = false;
 
     static final UncaughtExceptionHandler uncaughtExceptionHandler = (Thread t, Throwable e) -> {
         try {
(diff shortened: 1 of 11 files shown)


heroic-core/src/main/java/com/spotify/heroic/HeroicCore.java
```diff
@@ -139,6 +137,8 @@
 
     private static final int DEFAULT_LIGHTSTEP_REPORTING_MS = 1_000;
     private static final int DEFAULT_LIGHTSTEP_MAX_SPANS = 1_000;
+    static final boolean DEFAULT_GRPC_ROUNDROBIN = true;
```
hexedpackets: Looks like these variables can be private

heroic-core/src/main/java/com/spotify/heroic/HeroicCore.java
```diff
@@ -312,6 +312,11 @@ private void enableTracing(Map<String, Object> config) throws IOException {
...
             final Integer maxBufferedSpans = (Integer) config.getOrDefault("maxBufferedSpans",
                 DEFAULT_LIGHTSTEP_MAX_SPANS);
+            final Boolean grpcRoundRobin = (Boolean) config.getOrDefault("grpcRoundRobin",
+                DEFAULT_GRPC_ROUNDROBIN);
+            final Boolean grpcResetClient = (Boolean) config.getOrDefault("grpcResetClient",
+                DEFAULT_GRPC_RESETCLIENT);
+            final String grpcCollectorTarget = configMap.get("grpcCollectorTarget");
 
 
             if (collectorHost == null || accessToken == null) {
```
hexedpackets: if `grpcCollectorTarget` is required with no default, we should add a null check for it here

heroic-core/src/main/java/com/spotify/heroic/http/tracing/TextFormatGetter.java
```diff
@@ -35,6 +35,10 @@ public String get(final Object o, final String s) {
     final ContainerRequest request = (ContainerRequest) o;
     final List<String> requestHeader = request.getRequestHeader(s);
 
+    if (requestHeader == null) {
+      return null;
+    }
+
     if (requestHeader.size() > 0) {
```
hexedpackets: You could combine the two checks, `if (requestHeader != null && requestHeader.size() > 0)`

heroic-core/src/test/java/com/spotify/heroic/http/tracing/TextFormatGetterTest.java
```diff
@@ -68,4 +68,14 @@ public void testNullSpan () {
     }
   }
 
+  @Test(expected = SpanContextParseException.class)
+  public void testNullRequestHeader () throws Exception {
+
+    final List<String> headers = new ArrayList<>();
+
+    doReturn(headers).when(request).getRequestHeader("X-Cloud-Trace-Context");
```
hexedpackets: This test is the same as the one above it (`testNullSpan`). I think the intention is to test a null with `doReturn(null).when(...`

pom.xml
```diff
@@ -620,6 +612,13 @@
         <version>1.53.0</version>
       </dependency>
 
+      <dependency>
+        <groupId>com.google.auto.value</groupId>
+        <artifactId>auto-value</artifactId>
+        <version>1.5</version>
+        <scope>provided</scope>
```
hexedpackets: auto-value is already defined higher up in the pom with version 1.6.3. Does it need to be 1.5?

docs/heroic.example.yml
```diff
@@ -226,34 +226,25 @@ metadata:
...
-#  #  # @default "heroic"
-#  #  service: heroic
+#  #  GRPC round robin
+#  #  # @default true
+#  #  #grpcRoundRobin: true
+#  #  GRPC reset client
+#  #  # @default false
+#  #  #grpcResetClient: false
+#  #  GRPC Collector Target (required)
+#  #  #grpcCollectorTarget: "srv://_spotify-lightstep-satellite._grpc.services.gew1.spotify.net"
```
dmichel1: `srv://` can't be used in OSS because that is an internal grpc module we have for DNS records.
Instead we have to do something like `dns:///foo.googleapis`

https://github.com/grpc/grpc-java/blob/master/core/src/main/java/io/grpc/internal/DnsNameResolverProvider.java#L35

dmichel1: I also think you can add the port here `8282` and not need this in the builder for the grpc connection.

PR #421, opened 2019-02-28
Use kotlin for data classes in aggregation modules

hexedpackets:
This completely removes lombok from the simple and cardinality aggregation modules. Most of the data classes should compile to the same bytecode. I had to change an abstract class into an interface and a couple data classes into singletons, but those will behave the same with how they were used before.


aggregation/cardinality/pom.xml
```diff
@@ -16,12 +16,6 @@
   <name>Heroic: Cardinality Aggregations</name>
 
   <dependencies>
-    <dependency>
```
dmichel1: bye 🔥

aggregation/cardinality/src/main/java/com/spotify/heroic/aggregation/cardinality/ReduceHyperLogLogPlusCardinalityBucket.kt
```diff
@@ -0,0 +1,76 @@
...
+
+import java.io.IOException
+import java.util.concurrent.ConcurrentLinkedQueue
+
+/**
+ * Bucket that counts the number of seen events.
+ *
+ * @author udoprog
+ */
+data class ReduceHyperLogLogPlusCardinalityBucket(
```
dmichel1: Not related to this PR but would be great to document what all these different aggregations do. and if they are not used/useful, remove them?

PR #425, opened 2019-03-04
Remove lombok from the heroic-elasticsearch-utils module.

hexedpackets:


hexedpackets:
@dmichel1 Partially - there's a delombok that ships with lombok, I'm using that for the tostring and log4j replacements.

PR #429, opened 2019-03-12
Switch build/tests to use gradle instead of maven

hexedpackets:
The largest benefit of switching is to deal with the flaky failsafe/surefire runners that have been struggling to run the heroic-dist integration tests. Other benefits are flexible dependency management (I kept everything pinned to a specific version for now, but we can use partials if we chose to), proper handling of building/testing individual modules rather than the whole project, and much more readable build definitions.

Some notable changes from how it works under maven:
- findbugs is intentionally removed since travis excluded that profile, it would only ever be run if you manually ran it locally
- the checkstyle suppression now ignores all test files - the maven plugin apparently never ran against tests, I plan to make a separate PR to clean them up and enable checking for them
- protoc is downloaded by the gradle plugin so the system-wide installation was removed
- integration and unit tests are run together when doing `test` or `check`. it is possible to separate these into different tasks, but I wasn't sure how much value there is in doing so
- the `heroic-reporting` submodule was removed. this was only used to aggregate jacoco reports from other modules, which is now done as a task in the root gradle config
- dependencies between modules are explicit. only a few places have transitive declarations (the `api` ones), which makes it more obvious when a module is using a dependency
- the heroic-dist shaded jar runs as an application, which makes specifying the main class from the CLI unnecessary


gradlew.bat
```diff
@@ -0,0 +1,84 @@
+@if "%DEBUG%" == "" @echo off
```
dmichel1: are we supporting running heroic on windows now?

hexedpackets: Nah, its autogenerated by `gradle init`.

hexedpackets:
The `repackaged` repos I left as maven intentionally, those aren't actually part of the heroic project as they get installed to your local m2. I think we should consider pulling those out into separate repos, but in any case I didn't see much benefit to switching them to gradle.

gradlew.bat
dmichel1: is there value in having this file committed then?

PR #430, opened 2019-03-18
fix startup issues

dmichel1:
two commits to fix some issues when moving everything to kotlin.

@hexedpackets @lmuhlha


dmichel1:
@hexedpackets i rebased to get your gradle changes but looks like the IT grpc tests are still failing. Locally everything works.

PR #438, opened 2019-03-22
Move static assets from gh-pages branch.

hexedpackets:
Github lets us publish docs from the `/docs` folder instead of the `gh-pages` branch. Moving the documentation into master will reduce friction when we need to update docs along with a code change.


dmichel1:
nice! +1

Issue #454, opened 2019-04-08
Create a fault tree analysis for the pubsub consumers


hexedpackets:
Having a FTA documented would go a long way in helping to understand the failure modes of the consumers, and what appropriate SLOs are.

FTA diagram tool: https://scram-pra.org/

PR #489, opened 2019-06-04
Add consumer fault tree definition

hexedpackets:
Adding in the XML document for the PubSub consumer fault tree. This is just the input data to see if there are things we should adjust before displaying it nicely.

I need to adjust how it gets rendered before making a full documentation page for it - there are some issue like the text of the probabilities overrunning their space. I plan to have the documentation page show the interactive D3 graphed output, a link to this XML, and an explanation of the assumptions/usability of the fault tree.

Part of #454


hexedpackets:
The schema is based on https://raw.githubusercontent.com/rakhimov/scram/master/share/input.rng with a few small additions.

Issue #442, opened 2019-03-29
Update to Java 11


hexedpackets:
One of the blockers for Java > 8 is lombok which has numerous compatibility issues. Some work has already been done to replace lombok with straight java or kotlin (kotlin preferred for class generation). Fully replacing lombok should be considered a prereq for updating.

PR #492, opened 2019-06-07
[core,datastax] Delombokify a couple data classes and a sl4j helpers.

hexedpackets:
Some progress towards #442


heroic-core/src/main/java/com/spotify/heroic/common/CoreJavaxRestFramework.java
```diff
@@ -28,10 +28,13 @@
...
 import javax.ws.rs.container.ConnectionCallback;
 import javax.ws.rs.core.Response;
-import lombok.extern.slf4j.Slf4j;
+import org.slf4j.Logger;
 
-@Slf4j
 public final class CoreJavaxRestFramework implements JavaxRestFramework {
+
+    private static final Logger log =
+        org.slf4j.LoggerFactory.getLogger(CoreJavaxRestFramework.class);
```
dmichel1: For all these log statements you should be able to qualify them as `LoggerFactory.getLogger()`that way its shorter and also fits on 1 line

heroic-core/src/main/java/com/spotify/heroic/cluster/Update.kt
```diff
@@ -0,0 +1,105 @@
...
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.heroic.cluster
+
+import eu.toolchain.async.Transform
+import java.net.URI
+import java.util.*
```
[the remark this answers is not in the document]
hexedpackets: I haven't read too much on it, but my understanding is its because Kotlin has more than just classes namespaced, so doing something like importing a bunch of functions from a package can be tedious. Java style guides generally do the wildcards for static imports for the same reason.

heroic-core/src/main/java/com/spotify/heroic/common/CoreJavaxRestFramework.java
hexedpackets: Yeah, those were all generated by the Intellij delombok function. I prefer the qualified version but not enough to manually go through and handcraft all those lines.

PR #497, opened 2019-06-10
[docs] Update the configuration documentation

### LINKED ISSUE #440: De-duplicate the config documentation
Right now the example config is separated from both the documentation and the code. As I found in #439, some parts of it are potentially grossly outdated. Keeping both the docs and the config in sync with code changes seems unnecessary and leads to lots of duplication. Having the example config is very valuable, but perhaps we don't need the full descriptions everywhere. 

I suggest we keep just the docs updated with descriptions of what everything does, and trim the config to just the shape of the default parameters. There will still be some duplication, but I think that reduces it from what we have now while still giving a config people can just copy and uncomment to use.

@spotify/layer8

hexedpackets:
This is both a style and content update. All of the configuration options are documented. I trimmed down the example config to something more usable for a quickstart example, and removed the duplicate explanations of the options in the config.

The config parameters are documented next to their defaults now in one single file. My thinking is this makes it much easier to use as a reference, since there are links from a TOC at the top and you can use browser searching to find any given option without having to first go to the correct page for it. 

Closes #440


docs/_includes/federation-tail.html
```diff
@@ -3,7 +3,7 @@
   all known <em>shards</em> and merge the result.
 </p>
 
-<img style="width: 100%;" src="images/sharding.svg"></img>
+<img style="width: 100%;" src="{{ 'images/sharding.svg' | relative_url }}"></img>
```
dmichel1: what are these braces doing? `{{ }}` this some sort of template?

example/heroic.example.yml
```diff
@@ -1,287 +1,48 @@
...
 port: 8080
 
 # Cluster configuration.
 cluster:
-  # Local ID.
-  # @default Generated UUID.
-  #id: 88b3f090-49f8-43fe-b30e-5ce0e3889884
-  # When communicating with self, avoid using the network.
-  # @default false
-  #useLocal: false
```
dmichel1: is this a non-exist option?

example/heroic.example.yml
```diff
@@ -1,287 +1,48 @@
...
-    #     seeds:
-    #       - localhost:9200
-    #   # The number of writes this backend allows per second before rate-limiting kicks in.
-    #   # @default 3000
-    #   writesPerSecond: 3000
-    #   # The number of minutes a write will be cached for.
-    #   # @Default 240
-    #   writeCacheDurationMinutes: 240
+    # ElasticSearch-based metadata.
+    - type: elasticsearch
```
dmichel1: is there `memory` based backend for metadata and suggest? that would be the "quick start" way to get heroic up and running if there is and it works.

example/heroic.example.yml
```diff
@@ -1,287 +1,48 @@
...
-  # GRPC reset client
-  #grpcResetClient: false # @default false
+consumers:
+ - type: pubsub
+   # Topic and subscription will be created within the project if they don't exist,
+   # assuming Heroic has appropriate GCP permissions.
+   project: heroic
+   topic: metrics
+   subscription: heroic-consumer
+   schema: com.spotify.heroic.consumer.schemas.Spotify100
```
dmichel1: side thought - maybe we should remove the `Spotify` from the schema with the 1.0 release? There's nothing specific to Spotify in the schema and this would allow us to have a more standard spec. Would have to do the same thing in ffwd.

example/heroic.example.yml
```diff
@@ -1,287 +1,48 @@
...
+
+# Aggregation cache.
+cache:
+  backend:
+    # In-memory based aggregation cache.
+    type: memory
```
dmichel1: Is this enabled by default if this is present? Curious because we don't have this enabled internally.

docs/_includes/federation-tail.html
hexedpackets: Yeah Jekyll uses [liquid templates](https://jekyllrb.com/docs/liquid/).

hexedpackets: The `relative_url` prepends the full site path, since this site gets deployed in a sub-url on GitHub pages. So this ends up being `src="/heroic/images/sharding.svg"`

example/heroic.example.yml
hexedpackets: `useLocal`? That's an option, I documented it in the cluster section.

example/heroic.example.yml
hexedpackets: Nope, but `standalone` runs Elasticsearch in Heroic oddly enough. So it uses the full ES API, except to a single-node cluster on the same JVM.

example/heroic.example.yml
hexedpackets: Agreed, I'll make an issue for it. We can also remove the need for the fully qualified package so make it easier to configure.

hexedpackets: #499

example/heroic.example.yml
dmichel1: TIL

example/heroic.example.yml
hexedpackets: Nope, by default there is no caching. I'll update the `cache_backend` section to make that explicit.

example/heroic.example.yml
dmichel1: I'm not seeing `useLocal` in any of the files other than this one?

example/heroic.example.yml
dmichel1: tack!

example/heroic.example.yml
hexedpackets: Line 215 of config.md. Are you not expanding that file? Its big and collapsed by default, but its the most important file of this PR.

docs/content/_docs/config.md
```diff
@@ -0,0 +1,1022 @@
...
+
+# Enable distributed tracing of Heroic's operations.
+tracing: <tracing_config>
+```
+
+### [`<feature>`](#feature)
+
+Features are a way to modify the behaviour of the service. They are implemented as flags namespaced to Heroic.
+
+Features can be configured either on a per-query basis, or in the configuration section `features` to apply it to all queries by default. They can either be enabled or disabled. To enable a flag, you specify its name. To disable it, you specify it's name prefixed with a minux sign: `-<feature>`.
```
dmichel1: typo - /minux/minus

docs/content/_docs/config.md
```diff
@@ -0,0 +1,1022 @@
...
+# The port number that this node will bind to.
+port: <int> default = 9698
+
+# Frame size limit in bytes.
+maxFrameSize: <int> default = 10000000
+```
+
+#### [jvm](#jvm)
+
+Communicate directly between multiple JVMs running on the same host.
```
dmichel1: this is really only used for testing? maybe make a note of that?

docs/content/_docs/config.md
```diff
@@ -0,0 +1,1022 @@
...
+# reached, the result will be back-off so that another node in the cluster can be used instead.
+concurrentQueriesBackoff: <int>
+
+# How many fetches are allowed to be performed in parallel for each request.
+fetchParallelism: <int> default = 100
+
+# When true, any limits applied will be reported as a failure.
+failOnLimits: <bool> default = false
+
+# hreshold for defining a "small" query, measured in pre-aggregation sample size.
```
dmichel1: typo - `Threshold`

docs/content/_docs/config.md
```diff
@@ -0,0 +1,1022 @@
...
+# Initial nodes in the cluster to connect to. Any hosts without a port specified
+# are assumed to use port 9300. Useful to have masters that rarely change as seeds.
+seeds:
+  - <string> default = localhost:9300
+  ...
+
+# Dynamically sniff new nodes.
+sniff: <bool> default = false
+
+# How often to sniffpoll for new nodes when `sniff` is enabled.
```
dmichel1: /sniffpoll/poll/

example/heroic.example.yml
dmichel1: I see it now

docs/content/_docs/config.md
hexedpackets: Maybe we should deprecate it. The integration tests we run on it are identical to what we run over gRPC.

PR #508, opened 2019-06-21
[tracing] Access token is optional per satellite version 2019-06-12_16-06-52Z

lmuhlha:


hexedpackets:
I think it makes sense to leave it as an option in the configuration and just remove the throw. Passing `null` to `withAccessToken` should have the same behavior as not calling `withAccessToken`.

PR #514, opened 2019-07-10
[datastax] Remove legacy schema from datastax metric module.

### LINKED ISSUE #493: Deprecate datastax legacy schema
Other than some lombok refactoring, the legacy schema [hasn't been touched in > 2 years](https://github.com/spotify/heroic/tree/master/metric/datastax/src/main/java/com/spotify/heroic/metric/datastax/schema/legacy). Anyone using the Cassandra module should be using the nextgen schema instead, so let's deprecate and remove the legacy schema.

hexedpackets:
Closes #493


lmuhlha:
Should we do a check if anyone is using the legacy / did we do that already / do we not care?

hexedpackets:
Not sure how we could check, did you have something in mind?

Either way, we said we were doing this as 1.0. So if there is someone still using the legacy schema they will have to switch in order to upgrade Heroic.

lmuhlha:
I was thinking an announcement or something but that sounds like it's been long enough

hexedpackets:
Like an announcement in the docs? Is there someplace we could fit that in or are you thinking of a adding an "Updates" section?

PR #515, opened 2019-07-16
Metrics size

malish8632:
- Add batch size metrics by consumers


heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100Proto.java
```diff
@@ -104,6 +104,8 @@ public Consumer(
         ingestions.add(ingestion.write(new Ingestion.Request(s, MetricCollection.points(points))));
       }
 
+      reporter.reportMetricsIn(metrics.size());
```
hexedpackets: We should also count metrics coming in through the non-protobuf serializer, the `Spotify100` schema.

malish8632: Done

PR #516, opened 2019-07-16
Explicitly setting cell timestamp

malish8632:
Fixing cell timestamp to be used by BigTable built-in garbage collection.

### DIFF
--- a/metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/api/Mutations.java
+++ b/metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/api/Mutations.java
@@ -26,7 +26,9 @@ import com.google.common.collect.ImmutableList;
 import com.google.protobuf.ByteString;
 import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.TimeUnit;
 import lombok.Data;
+import org.joda.time.DateTime;
 
 public class Mutations {
     private final List<com.google.bigtable.v2.Mutation> mutations;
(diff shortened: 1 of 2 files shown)


hexedpackets:
I guess the timestamp column already exists and we just weren't using it, so should the amount of bytes stored stay the same? Or will this increase the size of the rows in BigTable?

metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/api/Mutations.java
```diff
@@ -66,6 +68,7 @@ public Builder setCell(
                     .newBuilder()
                     .setFamilyName(family)
                     .setColumnQualifier(columnQualifier)
+                    .setTimestampMicros(TimeUnit.MILLISECONDS.toMicros(DateTime.now().getMillis()))
```
dmichel1: Could also set this to `-1` to use the bigtable server time - https://cloud.google.com/bigtable/docs/reference/data/rpc/google.bigtable.v2#google.bigtable.v2.Mutation.SetCell

malish8632: good catch. changed

malish8632:
> I guess the timestamp column already exists and we just weren't using it, so should the amount of bytes stored stay the same? Or will this increase the size of the rows in BigTable?

I'm going to follow up on this one with Bigtable gurus. It is probably will increase the size but we need it to use built-in garbage collection.

PR #547, opened 2019-09-24
Update query_language.html

lmuhlha:


docs/content/_docs/query_language.html
```diff
@@ -284,12 +242,31 @@ <h4>Operators</h4>
...
   <tr>
     <th scope="row">RegEx On Tag</th>
     <td><pre><code class="language-hql">&lt;a&gt; ~ &lt;b&gt;</code></pre></td>
     <td><pre><code class="language-json">["~", &lt;a&gt;, &lt;b&gt;]</code></pre></td>
   </tr>
+  
+    <tr>
+    <th scope="row">Raw</th>
+    <td><pre><code class="language-hql">&lt;a&gt; q &lt;b&gt;</code></pre></td>
+    <td><pre><code class="language-json">["q", &lt;a&gt;, &lt;b&gt;]</code></pre></td>
```
hexedpackets: For `q`, isn't it a complex query and not a tag-operator? The examples here use `a` and `b` as tag placeholders, but I don't think that's what a query with `q` would look like.

docs/content/_docs/query_language.html
```diff
@@ -284,12 +242,31 @@ <h4>Operators</h4>
     <td><pre><code class="language-hql">&lt;a&gt; not in [&lt;b&gt;, ..]</code></pre></td>
     <td><pre><code class="language-json">["not", ["or", ["=", &lt;a&gt;, &lt;b&gt;], ..]]</code></pre></td>
   </tr>
+  
+  <tr>
+    <th scope="row">Tag True</th>
+    <td><pre><code class="language-hql">&lt;a&gt; true [&lt;b&gt;, ..]</code></pre></td>
```
hexedpackets: `true` and `false` don't seem to apply to tags, they're just static filters [which always return those values](https://github.com/spotify/heroic/blob/master/heroic-component/src/main/java/com/spotify/heroic/filter/TrueFilter.java#L47)

lmuhlha: was debating if i should document them or not in this case

docs/content/_docs/query_language.html
```diff
@@ -284,12 +242,30 @@ <h4>Operators</h4>
...
   </tr>
+  
+  <tr>
+    <th scope="row">Tag True</th>
+    <td>Returns True</td>
+  </tr>
 
+  <tr>
+    <th scope="row">Tag False</th>
+    <td>Returns False</td>
```
hexedpackets: We should still have the example here, it just doesn't apply to other filters/tags. So the false filter looks like just `[false]`

docs/content/_docs/query_language.html
```diff
@@ -244,52 +202,175 @@ <h4>Operators</h4>
...
+
+  <tr>
+    <th scope="row">Key Does Not Match</th>
+    <td><pre><code class="language-hql">$key != &lt;a&gt;</code></pre></td>
+    <td><pre><code class="language-json">["not", ["key", &lt;a&gt;]]</code></pre></td>
+  </tr>
+
+  <tr>
+    <th scope="row">Tag True</th>
+    <td><pre><code class="language-hql">[true]</code></pre></td>
```
hexedpackets: I'm pretty sure the HQL for true/false doesn't have the brackets around it. They also aren't tag related, they're just literal true and false. Titling it as `Tag True` would make me think its checking the truthiness of a tag.

docs/content/_docs/query_language.html
```diff
@@ -244,52 +202,175 @@ <h4>Operators</h4>
...
+  
+  <tr>
+    <th scope="row">RegEx On Tag</th>
+    <td><pre><code class="language-hql">&lt;a&gt; ~ &lt;b&gt;</code></pre></td>
+    <td><pre><code class="language-json">["~", &lt;a&gt;, &lt;b&gt;]</code></pre></td>
+  </tr>
+  
+  <tr>
+    <th scope="row">Raw</th>
+    <td>Custom filters</td>
```
hexedpackets: the `custom filters` cell messes up the table formatting

docs/content/_docs/query_language.html
```diff
@@ -244,52 +202,175 @@ <h4>Operators</h4>
...
+  <tr>
+    <th></th>
+    <th>Description</th>
+    <th>Parsed</th>
+    <th>JSON</th>
+  </tr>
+
+  <tr>
+    <th scope="row">Simple String</th>
+    <td>Simple string without quotes.</td>
```
hexedpackets: What's your thought on separating these from the Operators sections? Like what makes something an operator vs a custom filter? Some also seem to be duplicated between the two (like tag mataches).

docs/content/_docs/query_language.html
```diff
@@ -244,52 +202,175 @@ <h4>Operators</h4>
...
+    <td><pre><code class="language-hql">&lt;foo&gt; != &lt;bar&gt;</code></pre></td>
+    <td><pre><code class="language-json">["not", ["=", &lt;foo&gt;, &lt;bar&gt;]]</code></pre></td>
   </tr>
 
   <tr>
     <th scope="row">Key Matches</th>
-    <td><pre><code class="language-hql">$key = &lt;a&gt;</code></pre></td>
-    <td><pre><code class="language-json">["key", &lt;a&gt;]</code></pre></td>
+    <td>Key equals bar.</td>
+    <td><pre><code class="language-hql">$key = &lt;bar&gt;</code></pre></td>
```
hexedpackets: The `$key` key is only special by our convention, but its basically a normal tag and there's been some effort to move away from treating it as a special object. I think we can get rid of the sections related to it and just have the standard tag ones - eg the `Tag Matches` two rows up does the same thing as `Key Matches` if you give it `$key` as the tag.

lmuhlha:
@hexedpackets should be good to go, let me know what you think of the custom description

docs/content/_docs/query_language.html
```diff
@@ -227,72 +192,104 @@ <h4>Operators</h4>
...
+    <td><pre><code class="language-hql">&lt;foo&gt; !^ &lt;bar&gt;</code></pre></td>
+    <td><pre><code class="language-json">["not", ["^", &lt;foo&gt;, &lt;bar&gt;]]</code></pre></td>
   </tr>
 
   <tr>
-    <th scope="row">Tag Not In</th>
-    <td><pre><code class="language-hql">&lt;a&gt; not in [&lt;b&gt;, ..]</code></pre></td>
-    <td><pre><code class="language-json">["not", ["or", ["=", &lt;a&gt;, &lt;b&gt;], ..]]</code></pre></td>
+    <th scope="row">True</th>
+    <td><pre><code class="language-hql">true</code></pre></td>
```
hexedpackets: From this row down, there's no description cell which messes up the table format.

docs/content/_docs/query_language.html
```diff
@@ -227,72 +192,104 @@ <h4>Operators</h4>
...
 </table>
 </div>
 
+<h4 id="customfilters">
+  Custom Filters
+  <a class="link-to" href="#customfilters"><span class="glyphicon glyphicon-link"></span></a>
+</h4>
+
+<p>
+  When you need to select series using a more complex syntax than that available through the basic tag suggestions, you can create custom filters. These can take advantage of any of the filters mentioned above, or use the HQL language covered in the next section..
```
hexedpackets: Sounds good, I would add that it can only use the HQL format of the above filters.

Also looks like an extra trailing period .

docs/content/_docs/query_language.html
lmuhlha: oop. fixing

PR #557, opened 2019-10-04
[component] Final lombok removal.

hexedpackets:


lmuhlha:
seems like some checkstyle validation errors

PR #558, opened 2019-10-04
Take resource tags into account when generating the hash code

lmuhlha:
generateHash takes into account tags but not resource identifiers, so series with different resource identifiers end up looking the same. This leads to alerts not matching up and getting resolved properly, etc. 

Need to take the resource identifiers into account to remediate this issue, and keep a tag-only version for metadata/suggest cache matching.

### DIFF
--- a/heroic-core/src/main/java/com/spotify/heroic/shell/task/MetadataEntries.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/shell/task/MetadataEntries.java
@@ -97,7 +97,7 @@ public class MetadataEntries implements ShellTask {
                     io
                         .out()
                         .println(mapper.writeValueAsString(
-                            new AnalyticsSeries(series.getHashCode().toString(),
+                            new AnalyticsSeries(series.getHashCodeTagOnly().toString(),
                                 mapper.writeValueAsString(series))));
                 } catch (final Exception e) {
                     log.error("Failed to print series: {}", series, e);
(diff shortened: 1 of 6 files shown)


heroic-core/src/main/java/com/spotify/heroic/shell/task/MetadataEntries.java
```diff
@@ -97,7 +97,7 @@ public TaskParameters params() {
                     io
                         .out()
                         .println(mapper.writeValueAsString(
-                            new AnalyticsSeries(series.getHashCode().toString(),
+                            new AnalyticsSeries(series.getHashCodeTagOnly().toString(),
```
dmichel1: this means we will only have analytics around series that are within the metadata index meaning the analytics will be incomplete with regards to series that contain resource identifiers. perhaps this is okay for now and in the future we would want to think about tracking series that contain resource identifiers?  

wdyt @lmuhlha @hexedpackets

lmuhlha: They were already incomplete so I think it's okay to leave the behavior the same for now. We can iterate on it afterwards.

lmuhlha: We can also probably remove getHashCode now

lmuhlha: Removed it, gonna merge now 🎉

PR #559, opened 2019-10-07
Update dependencies to support Java 9+.

### LINKED ISSUE #442: Update to Java 11

hexedpackets:
Resolves #442

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: java
 
 jdk:
-  - openjdk8
+  - openjdk11
 
 branches:
   only:
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM maven:3.5.4-jdk-8 as builder
+FROM maven:3.5.4-jdk-11 as builder
 LABEL maintainer "layer8 <layer8@spotify.com>"
 
 COPY . .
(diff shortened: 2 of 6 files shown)


lmuhlha:
I think travis needs an update as well: https://github.com/spotify/heroic/blob/master/.travis.yml#L3

lmuhlha:
and the docs while we're here https://github.com/spotify/heroic/blob/f022d6f7e8eee66e5b2c23d0a0cd078cc5998664/docs/content/_docs/getting_started/compile.html#L17

lmuhlha:
possibly these two as well?
https://github.com/spotify/heroic/blob/master/build.gradle#L228
https://github.com/spotify/heroic/blob/master/build.gradle#L247

hexedpackets:
Good catches (can't respond inline since they weren't code comments). I updated travis and the docs.

jackson-datatype-jdk8 adds datatypes that didn't exist before Java 8, its needed for serializing/deserializing things like Optionals irregardless of using Java 11.

kotlin-stdlib-jdk8 is the same, but for interoperability between Java and Kotlin. Also works and is needed for Java 11.

PR #566, opened 2019-10-08
Deprecate events and remove the partial events implementation

lmuhlha:


lmuhlha:
docs/content/_api_types/event.md was removed in my other PR from earlier this week, but i'll fix the other 2 :)

Issue #554, opened 2019-10-03
Startup fails if suggest backend is unavailable


hexedpackets:
In spotify/heroic#406 the error checking was updated to not mark API nodes as failing if suggest is the only backend erroring. The healthchecking done when starting should use the same logic, right now startup will fail and the node will crash if suggest is down.

PR #568, opened 2019-10-10
[http] Add suggest backend to status API.

hexedpackets:


heroic-core/src/main/java/com/spotify/heroic/http/status/StatusResource.java
```diff
@@ -65,15 +73,21 @@ public StatusResource(
...
+        final StatusResponse.Backend suggestBackends = buildSuggestBackendStatus();
 
         final StatusResponse.Cluster cluster = buildClusterStatus();
 
         final boolean allOk =
-            consumers.isOk() && backends.isOk() && metadataBackends.isOk() && cluster.isOk();
+            consumers.isOk()
+            && backends.isOk()
+            && metadataBackends.isOk()
+            && suggestBackends.isOk()
```
lmuhlha: would removing this check help out with the restarts? or you're just separating stuff out first to test?

dmichel1: would adding suggest cause the status endpoint to fail if nots no longer `ok`? this end point is used in external health checks

hexedpackets: Ooh good point. I'll update this, maybe make it configurable what causes a failure.

hexedpackets: I dug into this more and this doesn't actually check the health of elasticsearch. For suggest, the `isOk` is passed through to the `isReady` method of the `Managed<Connection>` object. The `Managed` container creates and tears down the class in an async thread, only calling the constructor for the `Connection` class.

dmichel1: but if ES suggest was red wouldn't the `Connection` not become resolvable? I guess this is the behavior you couldn't reproduce in https://github.com/spotify/heroic/issues/554?

Issue #505, opened 2019-06-14
[tracing] Link spans properly


lmuhlha:
the http -> grpc spans are not linked. 
the grpc -> grpc spans are not linked (cross region spans)
the http -> grpc -> bigtable spans are not linked correctly.

Any leanings on how to pass context between highly async requests should be documented for others.

dmichel1:
After #615 the only think left is to link the http call to `CoreQueryManager`

dmichel1:
This has been completed as of #621

PR #569, opened 2019-10-10
Add tracing to the query path for bigtable

dmichel1:
Spans are not linked from the http thread pool nor linked to the gRPC calls. https://github.com/spotify/heroic/issues/505 will fix that

Finished product... Some of the spans for async operations like `localMetricsManager.findSeries` should be closed earlier but for some reason is not.

Screen Shot 2019-10-10 at 2 30 52 PM


heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -282,22 +306,32 @@ public QueryTrace buildTrace() {
                 }
 
                 final List<Callable<AsyncFuture<FetchData.Result>>> fetches = new ArrayList<>();
-
-                /* setup fetches */
                 accept(metricBackend -> {
                     for (final Series series : result.getSeries()) {
+                        final Span fetchSeries = tracer.spanBuilderWithExplicitParent(
```
hexedpackets: I think its a good pattern to end the spans in the same context where they're started. Closing it from the function where its passed in makes it easy to accidentally leak a span (especially if that child passes it further down the stack) and makes it harder to reason about what the span is actually measuring.

metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableBackend.java
```diff
@@ -413,46 +421,78 @@ public Statistics getStatistics() {
...
-            final AsyncFuture<List<FlatRow>> readRows = client.readRows(table, ReadRowsRequest
-                .builder()
-                .range(new RowRange(Optional.of(p.rowKeyStart), Optional.of(p.rowKeyEnd)))
-                .filter(RowFilter.chain(Arrays.asList(RowFilter
-                    .newColumnRangeBuilder(p.columnFamily)
-                    .startQualifierOpen(p.startQualifierOpen)
-                    .endQualifierClosed(p.endQualifierClosed)
-                    .build(), RowFilter.onlyLatestCell())))
-                .build());
+            Span readRowsSpan = tracer.spanBuilderWithExplicitParent(
```
hexedpackets: Same comment as above for closing out this span.

rpc/grpc/src/main/java/com/spotify/heroic/rpc/grpc/GrpcRpcProtocol.java
```diff
@@ -215,7 +216,8 @@ public ClusterNode node() {
             }
 
             @Override
-            public AsyncFuture<FullQuery> query(final FullQuery.Request request) {
+            public AsyncFuture<FullQuery> query(final FullQuery.Request request, final Span span) {
+                // TODO: Do we have to use a span here with gRPC since its auto instrumented?
```
hexedpackets: We don't need to setup a span here, but we should make sure the parent installed in the thread-local context when [the client starts its own span](https://github.com/spotify/heroic/blob/b5ae95cc520f9f33a16f1c9ef3a4e0722587d44f/rpc/grpc/src/main/java/com/spotify/heroic/rpc/grpc/GrpcRpcClient.java#L80). Than can be done either here, or by removing the span from the signature and ensuring its set wherever `query` is called. 

I'd vote for the second option for simplicity.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -340,12 +375,22 @@ public QueryTrace buildTrace() {
...
+            final Span metadataSpan = tracer.spanBuilderWithExplicitParent(
+                "localMetricsManager.findSeries", span).startSpan();
 
             return metadata
-                .findSeries(
-                    new FindSeries.Request(request.filter(), request.range(), seriesLimit))
+                .findSeries(new FindSeries.Request(request.filter(), request.range(), seriesLimit))
+                .onResolved(t -> metadataSpan.putAttribute(
+                    "seriesCount", longAttributeValue(t.getSeries().size())))
+                .onDone(new EndSpanFutureReporter(metadataSpan))
```
hexedpackets: If I'm reading the async source correctly, runnables on the future get executed in reverse order that they're set. Which would mean the transform is happening before metadataSpan ends, causing the incorrect timing you screenshotted. 

Check my math on that though: https://github.com/udoprog/tiny-async-java/blob/1.11.0/tiny-async-core/src/main/java/eu/toolchain/async/concurrent/ConcurrentResolvableFuture.java#L491

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
dmichel1: 👍 💯 - definitely that's a good best practice we should document.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
dmichel1: I moved the closing of the spans around, moving it to the bottom where `.onDone(reporter.reportQueryMetrics());` is produces...

Screen Shot 2019-10-16 at 10 26 51 AM

which makes the `findSeries` the entire length of the `fetch`


Adding it back to where it was actually shows some spans closing at the current timing

Screen Shot 2019-10-16 at 10 27 37 AM


vs others...

Screen Shot 2019-10-16 at 10 27 44 AM

very strange...

metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableBackend.java
dmichel1: fixed 👍

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -215,14 +233,20 @@ private Transform(
...
             public AsyncFuture<FullQuery> transform(final FindSeries result) throws Exception {
+                final Span fetchSpan = tracer.spanBuilderWithExplicitParent(
+                    "localMetricsManager.fetch", parentSpan).startSpan();
                 final ResultLimits limits;
 
                 if (result.getLimited()) {
                     if (failOnLimits) {
                         final RequestError error = new QueryError(
-                            "The number of series requested is more than the allowed limit of " +
+                            "The number of series requested is more than th e allowed limit of " +
```
hexedpackets: Typo?

dmichel1: fixed

PR #574, opened 2019-10-23
[aggregation] add ratePerSecond

### LINKED ISSUE #452: Support Rate/RatePerSecond aggregations
Heroic should support Rate and RatePerSecond (irate) aggregations. Heroic already has Delta and DeltaPerSecond aggregations that can be used as examples when implementing...

 

Calculating the rate will allow us to calculate and comprehend SLO's in a much clearer way since we are getting the "total number of failed requests" / "all requests" vs calculating this from rates that are already aggregated on instances. 

Its easier to say 100/100000 requests are failing vs 0.01/1.9 (calculated with the avg of the rates). Also summing up the requests produces NO precision lost or issues with low throughput services or newly created meters.

 

Prometheus descriptions...

https://github.com/prometheus/prometheus/blob/2bd510a63e48ac6bf4971d62199bdb1045c93f1a/docs/querying/functions.md#rate

https://github.com/prometheus/prometheus/blob/2bd510a63e48ac6bf4971d62199bdb1045c93f1a/docs/querying/functions.md#irate

How prometheus handles counter resets, https://github.com/prometheus/prometheus/blob/c6399296dcc059bbc496f40e31aa2c7b7a341951/promql/functions.go#L178

https://github.com/prometheus/prometheus/blob/c6399296dcc059bbc496f40e31aa2c7b7a341951/promql/testdata/functions.test#L94

dmichel1:
The rate per second aggregation calculates the rate per second between samples in a given extent.

Rate per second works with sampling. If a queries resolution is 1 minute and points are published every 30s each bucket will contain 2 points. The rate will be calculated by getting the difference between the two points values and dividing them by the time between the points.

Note: at least 2 points are required in order to calculate the rate. Otherwise the previous value is unknown.

It's intended to be used with cumulative counters. It will handle the counters being reset because of restarts, etc.

I reimplemented an earlier pull request #408 that implemented `ratePerSecond` similarly to how `deltaPerSecond` works which does not support sampling the data. I think sampling is a requirement for most aggregations otherwise a large number of timeseries and points will likely cause the frontend to crash or produce graphs that are hard to read.

This closes #452.

### DIFF
--- a/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/RatePerSecondInstance.kt
+++ b/aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/RatePerSecondInstance.kt
@@ -1,0 +1,89 @@
+/*
+ * Copyright (c) 2019 Spotify AB.
+ *
+ * Licensed to the Apache Software Foundation (ASF) under one
+ * or more contributor license agreements.  See the NOTICE file
+ * distributed with this work for additional information
+ * regarding copyright ownership.  The ASF licenses this file
+ * to you under the Apache License, Version 2.0 (the
+ * "License"); you may not use this file except in compliance
+ * with the License.  You may obtain a copy of the License at
+ *
+ *   http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.heroic.aggregation.simple
+
+import com.google.common.collect.ImmutableSet
+import com.google.common.util.concurrent.AtomicDouble
+import com.spotify.heroic.aggregation.AbstractBucket
+import com.spotify.heroic.aggregation.AggregationInstance
+import com.spotify.heroic.aggregation.BucketAggregationInstance
+import com.spotify.heroic.aggregation.DoubleBucket
+import com.spotify.heroic.metric.Metric
+import com.spotify.heroic.metric.MetricType
+import com.spotify.heroic.metric.Point
+
+/*
+    Calculate the rate per second of cumulative counters. When the counter is reset
+    because of an instance restart the interval will be skipped.
+ */
+data class RatePerSecondInstance(
+    override val size: Long,
+    override val extent: Long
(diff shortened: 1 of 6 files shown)


aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/RatePerSecondInstance.kt
```diff
@@ -0,0 +1,87 @@
...
+
+
+/*
+    Calculate the rate per second of cumulative counters. When the counter is reset
+    because of an instance restart the interval will be skipped.
+ */
+data class RatePerSecondInstance(
+    override val size: Long,
+    override val extent: Long
+) : BucketAggregationInstance<RatePerSecondInstance.RateBucket>(size,
```
dmichel1: @hexedpackets this is the part I read about the buckets needing to be thread safe https://github.com/spotify/heroic/blob/092c340c30460129840fb8511c825fd773822ca4/heroic-component/src/main/java/com/spotify/heroic/aggregation/BucketAggregationInstance.kt#L43

aggregation/simple/src/main/java/com/spotify/heroic/aggregation/simple/RatePerSecondInstance.kt
```diff
@@ -0,0 +1,87 @@
...
+
+    }
+
+    data class RateBucket(override val timestamp: Long) : AbstractBucket(), DoubleBucket {
+        private var firstPoint: Point? = null
+        private var lastPoint: Point? = null
+        private val rate = AtomicDouble()
+
+        override fun updatePoint(key: Map<String, String>, p: Point) {
+            if (firstPoint == null) {
```
dmichel1: and i'm pretttty sure this is not thread safe.

hexedpackets: Hmm yeah, this requires having chronologically serialized data.

dmichel1: i went through the calling code - when bigtable calls `metricsConsumer.accept(readResult);`  the entire metric collection is present and in order. 

As far as I can tell when doing a for-each aggregation, only a single thread is working on a given bucket at a time. I looked both at the code and also tracing what thread was executing the bucket (`grpc-default-executor-0` which is also weird? as that thread shouldn't be doing this work?).

dmichel1: @hexedpackets let's review this and confirm when you are back.

dmichel1:
I'm merging this after I rebased with the latest changes.

PR #581, opened 2019-11-15
Chain aggregation fix missing type id

lmuhlha:
Using:
 "aggregators": [{"type": "chain", "chain": [{"type": "delta"}, {"type": "sum", "size": "60s"}]}], 

Replicated the error in production. Tested the fix locally with the same query.


hexedpackets:
@lmuhlha Since this was a regression, I think we should add a chain deserialization test that would have caught the bug.

Issue #478, opened 2019-05-09
Analytics of deployments


hexedpackets:
We should start collecting analytics on where/how Heroic is deployed. This would need to be optional and very obvious to users when its enabled so they can easily opt-out. We'll also need to be very careful to only collect anonymous, high-level stats and nothing uniquely identifiable.

As part of this, we'll need somewhere to send the analytics to that's internet-reachable.

hexedpackets:
We should have a page in the docs detailing exactly what we collect. [Homebrew has a great example of this](https://docs.brew.sh/Analytics).

I think Google Analytics is a reasonably low-friction place to send the usage reports.

PR #582, opened 2019-11-19
[usagetracking] Create a module for tracking basic events.

hexedpackets:
Start of analytics support for #478 

TODOs on this:
- [x] get event showing up in google analytics console
- [x] fix checkstyle
- [x] add tests
- [x] document what we collect and how to disable it
- [x] add cluster size as an event parameter
---

Currently having problems with the event. I can see it firing in both the logs and the debugger, and Google Analytics returns a 200. I also checked that the event is valid by sending it to `https://www.google-analytics.com/debug/collect`. Despite this, nothing ever shows up in the Google Analytics console.

Any thoughts @spotify/observability ?


hexedpackets:
Figured it out with some help from @sjoeboo. Turns out there's a checkbox in the view settings for `Bot Filtering` which is enabled by default. Seems that anything that isn't a browser counts as a bot - at least, curl and okhttp do. Disabled that checkbox and events started showing up.

heroic-core/build.gradle
```diff
@@ -44,16 +45,20 @@ dependencies {
     implementation 'com.lightstep.tracer:tracer-okhttp'
     implementation 'com.lightstep.tracer:java-common'
     implementation 'com.lightstep.opencensus:lightstep-opencensus-exporter'
+
     implementation 'com.typesafe:config:1.3.2'
     implementation 'net.jodah:expiringmap:0.5.1'
+
+    // Default usage tracking module. Normally different module dependencies are handled in
+    // heroic-dist, but defaults need to be accessible in heroic-core.
+    implementation project(':heroic-usagetracking-google')
```
dmichel1: nit; can we call this `heroic-usage-tracking-google-analytics' or `heroic-usage-tracking-google'?

heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java
```diff
@@ -100,8 +103,9 @@ public CoreClusterManager(
       LocalClusterNode local,
       HeroicContext context,
       @Named("topology") Set<Map<String, String>> expectedTopology,
-
-      final QueryReporter reporter) {
+      final QueryReporter reporter,
+      @Named("cluster") UsageTracking usageTracking
```
dmichel1: why is this cluster? shouldn't it be `usageTracking`?

hexedpackets: It's the cluster version of the `UsageTracking` injected object, which distinguishes it from other Providers for `UsageTracking`.

heroic-core/build.gradle
hexedpackets: Sure

hexedpackets:
@dmichel1 good point. I'll merge this and add a scheduled event in a second PR

heroic-core/src/main/java/com/spotify/heroic/cluster/CoreClusterManager.java
hexedpackets: Actually, I realized that's not needed at all, I can just remove the Provides from the cluster module. Will update in the scheduled PR.

Issue #592, opened 2019-12-31
Upgrade elasticsearch client to 7.5


dmichel1:
ES 5.6 has been EOL since March 2019. The latest version is currently 7.5. Jumping two major versions is going to come with a number of breaking changes that need to be scoped out.

One such breaking change is that indexes no longer have types.

Heroic is currently using the transport client which will be removed in 8.x. Part of the scoping work should look at if its worth moving to the high level rest client now or at another time.

dmichel1:
I just got the metadata part of Heroic running with an ES 7.5 client.  Sticking with the transport client is the least amount of work. The suggest part of Heroic is going to be much more complicated.

dmichel1:
WIP branch here https://github.com/spotify/heroic/tree/dmichel/es7

PR #594, opened 2020-01-06
Upgrade elasticsearch client to 7.5

dmichel1:
Upgrade the ES client to 7.5 which is the latest version. This is the last major version that will support the transport client.

The major change from 5.x -> 7.x is indexes no longer support multiple types. 
* `suggest` has two types, series and tags. This means there will be 2x the number of indexes.
* `metadata` has only one type, metadata. There will be the same number of indexes.

Other changes made as part of the PR.
* Remove deprecated configuration options.
* Completely remove the StandaloneClient as it was removed in 6.x+. TestContainers are now used for the integration tests.
* Completely remove the NodeClient since it's not recommended to use anyway.

**WARNING** this is a breaking change and should be versioned as such. The 7.5x will not work with a 5.x cluster.

Closes https://github.com/spotify/heroic/issues/592.


suggest/elasticsearch/src/main/resources/com.spotify.heroic.suggest.elasticsearch/kv/series.json
```diff
@@ -1,8 +1,5 @@
 {
   "{{type}}": {
-    "_all": {
```
dmichel1: This is not a part of ES7 anymore.

PR #602, opened 2020-02-10
Deprecate / Remove groupBy field

lmuhlha:


docs/_includes/requests/post-query-metrics.json
```diff
@@ -1,5 +1,4 @@
 {
   "range": {"type": "relative", "unit": "HOURS", "value": 2},
-  "filter": ["and", ["key", "foo"], ["=", "foo", "bar"], ["+", "role"]],
-  "groupBy": ["site"]
```
hexedpackets: It would be helpful to add an example of the replacement aggregation here rather than just removing it.

lmuhlha: done 👍

PR #608, opened 2020-02-12
[metadata] clear search scroll when complete

dmichel1:
The search scroll must be closed when it's no longer needed otherwise elasticsearch will hold onto them until the timeout is reached. In previous versions of elasticsearch the number of scrolls that could be open was unlimited, with ES7 the default is now 500. It can be overriden with `search.max_open_scroll_context`.

https://www.elastic.co/guide/en/elasticsearch/reference/6.8/search-request-scroll.html#_clear_scroll_api

### DIFF
--- a/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/Connection.java
+++ b/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/Connection.java
@@ -29,7 +29,6 @@ import eu.toolchain.async.ResolvableFuture;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
-import javax.annotation.Nonnull;
 import org.elasticsearch.action.ActionListener;
 import org.elasticsearch.action.admin.indices.template.put.PutIndexTemplateRequestBuilder;
 import org.elasticsearch.action.delete.DeleteRequestBuilder;
@@ -142,7 +141,6 @@ public class Connection {
         return client.getClient().prepareSearchScroll(scrollId);
     }
 
-    @Nonnull
     public ClearScrollRequestBuilder clearSearchScroll(String scrollId) {
       return client.getClient().prepareClearScroll().addScrollId(scrollId);
     }
--- a/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/Connection.java
+++ b/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/Connection.java
@@ -29,6 +29,7 @@ import eu.toolchain.async.ResolvableFuture;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
+import javax.annotation.Nonnull;
 import org.elasticsearch.action.ActionListener;
 import org.elasticsearch.action.admin.indices.template.put.PutIndexTemplateRequestBuilder;
 import org.elasticsearch.action.delete.DeleteRequestBuilder;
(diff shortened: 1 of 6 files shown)


heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/Connection.java
```diff
@@ -140,6 +141,10 @@ public SearchScrollRequestBuilder prepareSearchScroll(String scrollId) {
         return client.getClient().prepareSearchScroll(scrollId);
     }
 
+    public ClearScrollRequestBuilder clearSearchScroll(String scrollId) {
```
hexedpackets: Is it a problem that `scrollId` is nullable (since `lastScrollId` in the `ScrollTransformResult` class is nullable)? What does the API do when you pass it null?

Either way, adding `@Nullable` to the method signature would be helpful to set expectations.

dmichel1: >What does the API do when you pass it null? 

resets the list of scrollId's to an empty array but the api will then fail the validation if the list is empty. Making this `@NotNull` instead and checking for null.

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/Connection.java
```diff
@@ -33,10 +33,12 @@
 import org.elasticsearch.action.admin.indices.template.put.PutIndexTemplateRequestBuilder;
 import org.elasticsearch.action.delete.DeleteRequestBuilder;
 import org.elasticsearch.action.index.IndexRequestBuilder;
+import org.elasticsearch.action.search.ClearScrollRequestBuilder;
 import org.elasticsearch.action.search.SearchRequestBuilder;
 import org.elasticsearch.action.search.SearchScrollRequestBuilder;
 import org.elasticsearch.action.support.master.AcknowledgedResponse;
 import org.elasticsearch.client.IndicesAdminClient;
+import org.jetbrains.annotations.NotNull;
```
dmichel1: this isn't what i wanted.

dmichel1:
@hexedpackets updated if you could PTAL.

malish8632:
How come we didn't see any issues when we mirrored traffic via envoy?

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/Connection.java
```diff
@@ -141,6 +142,7 @@ public SearchScrollRequestBuilder prepareSearchScroll(String scrollId) {
         return client.getClient().prepareSearchScroll(scrollId);
     }
 
+    @Nonnull
```
hexedpackets: Doesn't this annotate the return type rather than the argument?

I'm also not as concerned with the `Nonnull` annotation - I think our general approach has been to assume non-null, and do checking in places where its explicitly allowed (except for Kotlin classes, since its always explicit there).

dmichel1:
@malish8632 we only did 1% of production traffic when dark loading which wasn't enough to surface this issue.

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/Connection.java
dmichel1: then I can get rid of the annotation since I check for it before calling this method, right?

hexedpackets: yeah sgtm

dmichel1: 👍 removed

PR #609, opened 2020-02-12
Properly mark functions as deprecated

### LINKED ISSUE #603: Warn about deprecated code paths
We've tagged some functions/classes as `@Deprecated`. This only helps warn when trying to develop new features against that code path, however. Users of Heroic get no indication if they're using a config that loads a deprecated code path, and aren't likely to migrate unless that code breaks completely. We can improve this by logging a warning message.

lmuhlha:
Resolves #603


lmuhlha:
@hexedpackets I thiiiink this might accomplish what you were saying?

hexedpackets:
Its definitely good to mark the methods in addition to the Java docs, but won't this still just generate compilation warnings?

lmuhlha:
"Using the @Deprecated annotation to deprecate a class, method, or field ensures that all compilers will issue warnings when code uses that program element. In contrast, there is no guarantee that all compilers will always issue warnings based on the @deprecated Javadoc tag, though the Sun compilers currently do so. Other compilers may not issue such warnings. Thus, using the @Deprecated annotation to generate warnings is more portable that relying on the @deprecated Javadoc tag.

In addition, the @Deprecated annotation causes the javadoc-generated documentation to be marked "Deprecated" wherever that program element appears.

NOTE: Deprecation applies to classes and to individual methods or properties, not to their names. It is possible for a single method to have deprecated and non-deprecated overloadings. It is possible for a non-deprecated property to hide or override a deprecated one, which removes deprecation. As developer of an API, it is your responsibility to deprecate overrides of a deprecated method, if in fact they should be deprecated."

Yeah I guess it is still just compiler bleh

PR #615, opened 2020-02-25
[tracing] improve tags and refactor query spans

dmichel1:
* tracing config allows an arbitrary list of tags to be applied to incoming request spans.
* GCP platform tags are added to incoming request spans.
* query spans have been improved to link correctly to the child spans.
  * the bigtable grpc readRows span is now scoped and linked correctly.
  * the outgoing grpc request to different heroic shards is now scoped and linked correctly.
  * span will include if the response from bigtable contained resource identifiers. could be useful when tracking down latency issues.

Related to https://github.com/spotify/heroic/issues/505

Screen Shot 2020-02-25 at 2 24 14 PM


heroic-core/src/main/java/com/spotify/heroic/http/query/QueryResource.java
```diff
@@ -136,11 +136,6 @@ public void metrics(
 
         query.getQueries().ifPresent(queries -> {
             for (final Map.Entry<String, QueryMetrics> e : queries.entrySet()) {
-
```
dmichel1: this wasn't useful so i removed it

heroic-component/src/main/java/com/spotify/heroic/tracing/EnvironmentMetadata.java
```diff
@@ -0,0 +1,88 @@
...
+
+        attributes.put("gcp.project", stringAttributeValue(GcpMetadataLookup.getProjectId()));
+        String zone = GcpMetadataLookup.getZone();
+        attributes.put("gcp.zone", stringAttributeValue(zone));
+        attributes.put(
+            "gcp.region", stringAttributeValue(zone.substring(0, zone.lastIndexOf("-"))));
+    }
+
+    private static void addHostname() {
+        final Optional<String> hostname = getEnvVariable("HOSTNAME");
```
hexedpackets: I think it makes more sense to have this as part of the tracing config so everything can continue to be set in the heroic config file.

That might require using dagger instead of directly creating a singleton. I haven't fully thought through the instantiation but it feels more in-line with the rest of Heroic/tracing config loading.

heroic-core/src/main/java/com/spotify/heroic/http/tracing/OpenCensusApplicationEventListener.java
```diff
@@ -25,6 +25,7 @@
 import static java.text.MessageFormat.format;
 
 import com.spotify.heroic.common.ServiceInfo;
+import com.spotify.heroic.tracing.EnvironmentMetadata;
```
hexedpackets: Unused import - probably why the build is failing

heroic-component/src/main/java/com/spotify/heroic/tracing/TracingMetadataProvider.java
```diff
@@ -0,0 +1,29 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.heroic.tracing;
+
+import io.opencensus.trace.AttributeValue;
+import java.util.Map;
+
+public interface TracingMetadataProvider {
```
hexedpackets: I don't think this a useful interface. There's one implementation, and the places that use it refer directly to to that implementation.

heroic-component/src/main/java/com/spotify/heroic/tracing/GcpMetadataLookup.java
```diff
@@ -0,0 +1,77 @@
...
+
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
+import java.net.HttpURLConnection;
+import java.net.URL;
+
+/**
+ * Class copied from version 0.107.0 of
```
hexedpackets: We have google-cloud-core as a dependency already, we could use the MetadataConfig directly from that instead of copying.

heroic-component/src/main/java/com/spotify/heroic/tracing/RandomParentSampler.kt
```diff
@@ -0,0 +1,44 @@
+package com.spotify.heroic.tracing
```
hexedpackets: Missing license header - I'm not sure if the license check runs for kotlin files.

heroic-component/src/main/java/com/spotify/heroic/tracing/RandomParentSampler.kt
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * This will cause traces to not be complete and will likely not capture the outliers! It can
+ * still be useful to use this sampler when you have potentially tens of thousands of spans being
+ * created, as this many spans cannot be shown in any meaningful way within the current UX of
+ * many tracing solutions.
+ */
+class RandomParentSampler(private val probability: Double) : Sampler() {
+    init {
+        if (probability < 0.0 || probability > 1.0) {
+            throw IllegalArgumentException("probability must be in range [0.0, 1.0]")
```
hexedpackets: Might want to put a `@Throws` annotation on the constructor for usage in Java classes.

heroic-component/src/main/java/com/spotify/heroic/tracing/RandomParentSampler.kt
```diff
@@ -0,0 +1,44 @@
...
+    }
+
+    override fun shouldSample(
+        parentContext: SpanContext?,
+        hasRemoteParent: Boolean?,
+        traceId: TraceId?,
+        spanId: SpanId?, name: String?,
+        parentLinks: MutableList<Span>?): Boolean {
+
+        // Keep propagating parent sampling downstream.
```
hexedpackets: This comment is misleading - parent sampling decisions are not propagated downstream. The default is for isSampled is `false` so the logic is really:
- parent was sampled? apply probability to child
- no parent decision? explicitly do not sample

heroic-component/src/test/java/com/spotify/heroic/tracing/RandomParentSamplerTest.kt
```diff
@@ -0,0 +1,44 @@
...
+import kotlin.test.assertFailsWith
+import kotlin.test.assertFalse
+import kotlin.test.assertTrue
+
+class RandomParentSamplerTest {
+
+    lateinit var sampler: Sampler
+    @Before
+    fun setup() {
+         sampler = RandomParentSampler(1.0)
```
hexedpackets: For useful testing I think you would have to pass in the random generator as a sampler argument, letting you mock it here. I'm undecided about whether that's worth doing.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -199,6 +201,9 @@ public String toString() {
             private final Span parentSpan;
             private final MetricType source;
 
+            private static final double PROBABILITY = 0.005;
```
hexedpackets: This should definitely be configurable - possibly using the global probability as the default and having a config option to override it.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -311,8 +316,16 @@ public QueryTrace buildTrace() {
                 final List<Callable<AsyncFuture<FetchData.Result>>> fetches = new ArrayList<>();
                 accept(metricBackend -> {
                     for (final Series series : result.getSeries()) {
-                        final Span fetchSeries = tracer.spanBuilderWithExplicitParent(
-                            "localMetricsManager.fetchSeries", fetchSpan).startSpan();
+                        // Without sampling spans this would produce way too many.
+                        final SpanBuilder fetchSeriesSpanBuilder =
+                            tracer.spanBuilderWithExplicitParent(
+                                "localMetricsManager.fetchSeries", fetchSpan);
+                        if (result.getSeries().size() > 100) {
```
hexedpackets: The series limit should probably also be configurable, but if not it should at least be a class static.

hexedpackets: You could also move the `result.getSeries().size()` above the for loop to avoid counting the list every iteration.

metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableBackend.java
```diff
@@ -447,30 +449,37 @@ public Statistics getStatistics() {
...
+                                            RowFilter.newColumnRangeBuilder(p.columnFamily)
+                                                .startQualifierOpen(p.startQualifierOpen)
+                                                .endQualifierClosed(p.endQualifierClosed)
+                                                .build(),
+                                            RowFilter.onlyLatestCell())))
+                                .build())
+                        .onDone(new EndSpanFutureReporter(readRowsSpan));
+            }
 
+            final AtomicBoolean foundResourceIdentifier = new AtomicBoolean(false);
```
hexedpackets: Does this need to be an AtomicBoolean instead of a plain Boolean? Its threaded access, but it can only every get flipped to `true` so should be fine to use across threads. Not sure if it matters/has performance impact.

heroic-component/src/main/java/com/spotify/heroic/tracing/EnvVariableLookup.java
```diff
@@ -0,0 +1,31 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing,
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied.  See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.heroic.tracing;
```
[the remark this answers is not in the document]
dmichel1: 👍 - I can refactor right after we get this merged.

heroic-core/src/main/java/com/spotify/heroic/http/tracing/OpenCensusApplicationEventListener.java
dmichel1: this is from refactoring this module - it's used below `private static final EnvironmentMetadata environmentMetadata = EnvironmentMetadata.create();`

rpc/grpc/build.gradle
```diff
@@ -2,6 +2,7 @@ apply plugin: 'kotlin-kapt'
 
 dependencies {
     implementation project(':heroic-component')
+    implementation project(':heroic-core')
```
dmichel1: @hexedpackets when I moved  the `EnvironmentMetadata` to heroic-core i had to add this dependency which seems a bit heavy-handed.

heroic-component/src/main/java/com/spotify/heroic/tracing/EnvironmentMetadata.java
hexedpackets: I like the idea of just moving this to the tracing config as a list of tags, and setting the hostname tag could be done through config management. I do think it needs to be configurable - neither the env var HOSTNAME nor `gethostname` calls necessarily return a meaningful global identifier, and having to configure those just for the purpose of this tag doesn't seem great.

Example: on my local system, HOSTNAME is not set and `gethostname` returns a single non-domained host. In k8s, HOSTNAME is set but both it and `gethostname` are just the pod name without namespace/cluster/region/etc.

rpc/grpc/build.gradle
hexedpackets: Hmm yeah. Before this the dependencies on core are dist and test. I think there's some larger refactoring needed, for now leave it whichever way you prefer (in core or in component).

hexedpackets:
Tests still failing on checkstyle in btw

heroic-component/src/main/java/com/spotify/heroic/tracing/EnvironmentMetadata.java
dmichel1: you can now set tags via the config.

dmichel1:
@hexedpackets i'm ready for another review.

heroic-component/src/main/java/com/spotify/heroic/tracing/TracingConfig.kt
```diff
@@ -4,6 +4,7 @@ data class TracingConfig (
     val probability: Double = 0.01,
     val zpagesPort: Int = 0,
     val requestHeadersToTags: List<String> = listOf(),
+    val tags: Map<String, String> = mutableMapOf(),
```
hexedpackets: I don't think this needs to be mutable

dmichel1: 👍 changed to a read-only map.

PR #621, opened 2020-03-12
Scope request spans so that they link to coreQueryManager.

hexedpackets:
I believe this is the final part to close out #505. I also fixed the `findSeries` span so that its end time encompasses the `fetch` child spans.


heroic-core/src/main/java/com/spotify/heroic/http/tracing/OpenCensusApplicationEventListener.java
```diff
@@ -188,9 +186,10 @@ public void onEvent(RequestEvent event) {
                         invocable.getDefinitionMethod()));
 
                     final String spanName = invocable.getHandler().getHandlerClass().getName();
-                    resourceSpan = globalTracer
+                    globalTracer
```
dmichel1: 👏

PR #623, opened 2020-03-17
[elasticsearch] Support any ES index settings.

### LINKED ISSUE #622: Support additional ES index settings
When elasticsearch (either metadata or suggest) is enabled with `configure: true`, Heroic will write the expected mappings to as an index template. Ideally, using the built-in `configure` should provide all the configuration needed for a dedicated ES cluster. At a minimum this means adding replica/shard settings to the template.

hexedpackets:
Any arbitrary [index settings](https://www.elastic.co/guide/en/elasticsearch/reference/current/index-modules.html#index-modules-settings) can now be applied as part of the template (rotating index) or index creation (single index). Closes #622


heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/index/SingleIndexMapping.java
```diff
@@ -24,34 +24,42 @@
 import com.fasterxml.jackson.annotation.JsonCreator;
 import com.fasterxml.jackson.annotation.JsonProperty;
 import com.google.common.collect.ImmutableList;
+import java.util.HashMap;
```
dmichel1: wdyt about making this a kotlin data class too given `RotatingIndexMapping` was converted?

hexedpackets: sure, I initially wasn't touching it and then adding the settings in last minute. I'll convert it

hexedpackets:
> one comment but LGTM - this will allow us to use the shell with the proper settings in the config +1
> 
> IT tests are failing

Yeah they're real failures, need to figure out what broke.

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/Connection.java
```diff
@@ -82,36 +87,37 @@ public Connection(final AsyncFramework async, final IndexMapping index,
...
-          final String s = index.template().replaceAll("\\*", indexType + "-*");
-
-          log.info("[{}] updating template for {}", templateWithType, s);
+            Map<String, Object> settings = new HashMap<>();
+            settings.put("index", index.settings());
 
             final PutIndexTemplateRequestBuilder put = indices.preparePutTemplate(templateWithType)
-                .setSettings(type.getSettings())
-                .setPatterns(List.of(s))
+                .setSettings(settings)
```
dmichel1: these settings need to be merged with `type.getSettings` i believe.

PR #624, opened 2020-03-18
[suggest-elasticsearch] write series and tags in bulk

dmichel1:
bulk writes are supposed to reduce overhead and increase indexing speed.

https://www.elastic.co/guide/en/elasticsearch/reference/current/docs-bulk.html

### DIFF
--- a/heroic-test/src/main/java/com/spotify/heroic/test/AbstractSuggestBackendIT.java
+++ b/heroic-test/src/main/java/com/spotify/heroic/test/AbstractSuggestBackendIT.java
@@ -21,11 +21,13 @@
 
 package com.spotify.heroic.test;
 
+import static com.spotify.heroic.filter.Filter.matchKey;
+import static org.junit.Assert.assertEquals;
+
 import com.google.common.collect.ImmutableList;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.collect.ImmutableSet;
 import com.google.common.collect.ImmutableSortedSet;
-
 import com.spotify.heroic.HeroicConfig;
 import com.spotify.heroic.HeroicCore;
 import com.spotify.heroic.HeroicCoreInstance;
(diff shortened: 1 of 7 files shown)


suggest/elasticsearch/src/main/java/com/spotify/heroic/suggest/elasticsearch/SuggestBackendKV.java
```diff
@@ -21,17 +21,11 @@
```
dmichel1: my formatter kept messing these lines up... my intelij is set to use the google style guide

@hexedpackets which one do we use in heroic?

https://github.com/spotify/heroic/pull/624/files?w=1 shows this file without all the whitespace changes

suggest/elasticsearch/src/main/java/com/spotify/heroic/suggest/elasticsearch/SuggestBackendKV.java
```diff
@@ -645,204 +656,221 @@ public boolean isReady() {
...
+          bulkRequestBuilder.execute(bind(future));
+
+          return future
+              .directTransform(
+                  response -> {
+                    boolean addFailureAnnotation = true;
+                    final List<RequestError> errors = new ArrayList<>();
+                    for (final BulkItemResponse r : response.getItems()) {
+                      if (r.isFailed()) {
+                        errors.add(NodeError.internalError(r.getFailureMessage()));
```
dmichel1: With the way `RequestError` is designed a `NodeError` here is easiest error to use without modifying the response API's dramatically.

heroic-test/src/main/java/com/spotify/heroic/test/AbstractSuggestBackendIT.java
```diff
@@ -146,6 +150,17 @@ public final void abstractTeardown() throws Exception {
         core.shutdown().get();
     }
 
+    @Test
+    public void writeDuplicatesReturnErrorInResponse() throws Exception {
```
hexedpackets: Since this is only relevant to SuggestBackendKVIT, it should be moved into that class instead of overriding it with a blank test in MemoryBackendIT.

suggest/elasticsearch/src/main/java/com/spotify/heroic/suggest/elasticsearch/SuggestBackendKV.java
hexedpackets: The style XML for intellij is in the repo - https://github.com/spotify/heroic/blob/master/idea/code-style.xml

heroic-test/src/main/java/com/spotify/heroic/test/AbstractSuggestBackendIT.java
dmichel1: I tried moving it there but ran into timing issues brining up the elasticsearch test container. let me see if I can work past that...

suggest/elasticsearch/src/main/java/com/spotify/heroic/suggest/elasticsearch/SuggestBackendKV.java
dmichel1: weird - i think the reformatting must have just been really off in this file -- reformatting with the heroic style didn't change many things back.

PR #628, opened 2020-03-31
[elasticsearch] Connect with REST client

### LINKED ISSUE #618: ES data nodes DNS cache on Heroic causing issues when ES nodes fail
The solution should purge failed data nodes from the cached DNS list.

hexedpackets:
This adds the REST client as an option for querying ElasticSearch, which will become a hard requirement in 8.x. The REST client handles cluster sniffing and load balancing differently than the transport client; using this should close #618. 

I also removed references to deprecated ES types in this PR. Since all types were already in their own index, I don't think this will have an impact, but upgrading will probably require switching the API nodes before the consumers.


heroic-test/src/test/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKVTransportIT.java
```diff
@@ -0,0 +1,48 @@
...
+import com.spotify.heroic.elasticsearch.TransportClientWrapper;
+import com.spotify.heroic.elasticsearch.index.RotatingIndexMapping;
+import com.spotify.heroic.metadata.MetadataModule;
+import com.spotify.heroic.test.AbstractMetadataBackendIT;
+import com.spotify.heroic.test.ElasticSearchTestContainer;
+import java.util.List;
+import java.util.UUID;
+
+public class MetadataBackendKVTransportIT extends AbstractMetadataBackendKVIT {
+
```
lmuhlha, line 35: just a nitpick about extra space here

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/RestClientWrapper.kt
```diff
@@ -0,0 +1,81 @@
...
+import java.net.InetAddress
+import java.net.UnknownHostException
+import java.util.concurrent.TimeUnit
+
+private const val DEFAULT_PORT = 9200
+
+data class RestClientWrapper @JvmOverloads constructor(
+    val seeds: List<String> = listOf("localhost"),
+    val sniff: Boolean = false,
+    val sniffInterval: Duration = Duration.of(5, TimeUnit.MINUTES)
```
dmichel1, line 41: this is a high default -- iirc the default in the transport client was 30 seconds. 

make this at least 1 minute?

docs/content/_docs/config.md
```diff
@@ -624,9 +624,31 @@ settings: <map>
 
 The Elasticsearch client configuration to use.
 
+###### rest
+
+Connect using the High Level REST client. This is the preferred client for Elasticsearch and a replacement for the older transport client. This is compatabile with Elasticsearch node versions >= 7.5 and < 8.0.
```
dmichel1, line 629: the high level rest client does not work with >= 8.0 cluster?

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/RestClientWrapper.kt
hexedpackets: Oops missed this before merging. This is the same default as the transport library, but yeah we set the default in Heroic to 30s. I'll make a new PR to update it.

docs/content/_docs/config.md
hexedpackets: This specific version is not guaranteed to be 100% compatible with an 8.x cluster. The REST client guarantees forward compatibility of all minor versions, but not majors.

PR #639, opened 2020-04-23
[metadata] add metric for failed shards

### LINKED ISSUE #633: Add more metrics along the query path
There should be more than just async operation related metrics along the query path.

For example, when we ask the elasticsearch metadata backend for what series exist there should be metrics around whether or not the response contains any failure such as failed shards.

dmichel1:
The `failed-es-shards` metric can be used to track how often an elasticsearch shard fails.

Ideally this metric would be added to an elasticsearch specific metric reporter but that framework doesn't currently exist.

Closes #633

### DIFF
--- a/statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticMetadataBackendReporter.java
+++ b/statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticMetadataBackendReporter.java
@@ -57,6 +57,7 @@ public class SemanticMetadataBackendReporter implements MetadataBackendReporter 
     private final Counter entries;
     private final Counter writesDroppedByCacheHit;
     private final Counter writesDroppedByDuplicate;
+    private final Counter failedShards;
 
 
     public SemanticMetadataBackendReporter(SemanticMetricRegistry registry) {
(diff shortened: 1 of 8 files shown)


dmichel1:
I added a unit test for the metric being incremented when there is a failure.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticMetadataBackendReporter.java
```diff
@@ -74,17 +75,21 @@ public SemanticMetadataBackendReporter(SemanticMetricRegistry registry) {
...
         entries = registry.counter(base.tagged("what", "entries", "unit", Units.COUNT));
 
         writesDroppedByCacheHit = registry.counter(
             base.tagged("what", "writes-dropped-by-cache-hit", "unit", Units.COUNT));
         writesDroppedByDuplicate = registry.counter(
             base.tagged("what", "writes-dropped-by-duplicate", "unit", Units.COUNT));
 
+        // only relevant to es backend.
+        failedShards = registry.counter(
+            base.tagged("what", "failed-es-shards", "unit", Units.COUNT));
```
hexedpackets, line 92: Since this is only for metadata, can we add a tag for that - `component=metadata` or something?

dmichel1: @hexedpackets that's already done above on line 64. 

` final MetricId base = MetricId.build().tagged("component", COMPONENT);`

PR #642, opened 2020-04-23
[elasticsearch-metadata] Add option for live cursor

### LINKED ISSUE #636: Replace ElasticSearch scrolling
Scrolling in ES is [not designed for real-time user requests](https://www.elastic.co/guide/en/elasticsearch/reference/7.6/search-request-body.html#request-body-search-scroll) and has lead to ES related incidents. Instead of scrolling, we could try using the `search_after` request parameter to get a [live cursor](https://www.elastic.co/guide/en/elasticsearch/reference/current/search-request-body.html#request-body-search-search-after).

hexedpackets:
This adds a new feature, `com.spotify.heroic.metadata.live_cursor`, which changes the behavior of Elasticsearch searches for metadata. When enabled the results will be paginated with a live cursor instead of a snapshotted scroll.

Most likely we want this to be the default behavior, but its disabled by default so we can roll it out and test further.

Closes #636

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/metadata/FindSeries.kt
+++ b/heroic-component/src/main/java/com/spotify/heroic/metadata/FindSeries.kt
@@ -24,9 +24,11 @@ package com.spotify.heroic.metadata
 import com.fasterxml.jackson.annotation.JsonIgnore
 import com.spotify.heroic.cluster.ClusterShard
 import com.spotify.heroic.common.DateRange
+import com.spotify.heroic.common.Features
 import com.spotify.heroic.common.OptionalLimit
 import com.spotify.heroic.common.Series
 import com.spotify.heroic.filter.Filter
+import com.spotify.heroic.metric.FullQuery
 import com.spotify.heroic.metric.RequestError
 import com.spotify.heroic.metric.ShardError
 import eu.toolchain.async.Collector
@@ -73,6 +75,14 @@ data class FindSeries(
     data class Request(
         val filter: Filter,
         val range: DateRange,
-        val limit: OptionalLimit
-    )
+        val limit: OptionalLimit,
+        val features: Features
+    ) {
+        companion object {
+            @JvmStatic
+            fun withLimit(request: FullQuery.Request, limit: OptionalLimit): Request {
+                return Request(request.filter(), request.range(), limit, request.features())
+            }
+        }
+    }
 }
--- a/heroic-core/src/main/java/com/spotify/heroic/http/metadata/MetadataResource.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/http/metadata/MetadataResource.java
@@ -25,6 +25,7 @@ import com.google.common.collect.ImmutableList;
 import com.spotify.heroic.QueryDateRange;
 import com.spotify.heroic.QueryManager;
 import com.spotify.heroic.common.DateRange;
+import com.spotify.heroic.common.Features;
 import com.spotify.heroic.common.JavaxRestFramework;
 import com.spotify.heroic.common.OptionalLimit;
 import com.spotify.heroic.common.Series;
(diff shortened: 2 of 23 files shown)


heroic-core/src/main/java/com/spotify/heroic/http/metadata/MetadataResource.java
```diff
@@ -113,12 +114,16 @@ public void addSeries(@Suspended final AsyncResponse response, final Series seri
...
+            request::getFilter,
+            request::getRange,
+            () -> OptionalLimit.of(request.getLimit().orElse(MetadataQueryBody.DEFAULT_LIMIT))
+        );
 
-        httpAsync.bind(response, query
-            .useDefaultGroup()
-            .findSeries(new FindSeries.Request(c.getFilter(), c.getRange(), c.getLimit())));
+        FindSeries.Request seriesRequest = new FindSeries.Request(
+            criteria.getFilter(), criteria.getRange(), criteria.getLimit(), Features.DEFAULT);
```
dmichel1, line 124: this is effectively hardcoding this endpoint to the default features right? i think we only care about testing the metric path anyway

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/AbstractElasticsearchMetadataBackend.java
```diff
@@ -103,27 +131,32 @@ public AbstractElasticsearchMetadataBackend(final AsyncFramework async, final St
...
+        final Boolean scrolling;
 
-        @java.beans.ConstructorProperties({ "async", "limit", "converter", "scrollFactory" })
-        public ScrollTransform(
+        public SearchTransform(
             final AsyncFramework async,
             final OptionalLimit limit,
             final Function<SearchHit, T> converter,
-            final Function<String, Supplier<AsyncFuture<SearchResponse>>> scrollFactory
+            final BiFunction<String, SearchHit, Supplier<AsyncFuture<SearchResponse>>>
```
dmichel1, line 141: nit; this is a monster of a parameter... is this the way the formatter formatted it? 

is this more readable?
```
final BiFunction
    <String, SearchHit, Supplier<AsyncFuture<SearchResponse>>> searchFactory
```

heroic-component/src/main/java/com/spotify/heroic/metadata/FindSeries.kt
```diff
@@ -73,6 +75,14 @@ data class FindSeries(
...
         val filter: Filter,
         val range: DateRange,
-        val limit: OptionalLimit
-    )
+        val limit: OptionalLimit,
+        val features: Features
+    ) {
+        companion object {
+            @JvmStatic
+            fun withLimit(request: FullQuery.Request, limit: OptionalLimit): Request {
```
dmichel1, line 83: i like this factory 👍

metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
```diff
@@ -437,36 +433,42 @@ public boolean isReady() {
...
             modifier.accept(request);
 
-            AsyncFuture<ScrollTransformResult<T>> scroll = scrollEntries(
-                c, request, limit, converter);
-
-            return scroll
-                .onResolved(r -> ofNullable(r.getLastScrollId())
-                    .ifPresent(c::clearSearchScroll))
-                .directTransform(collector);
+            if (seriesRequest.getFeatures().hasFeature(Feature.METADATA_LIVE_CURSOR)) {
```
dmichel1, line 462: shouldn't this be `!seriesRequest.getFeatures().hasFeature(Feature.METADATA_LIVE_CURSOR)` ? 

otherwise this will use scroll when feature METADATA_LIVE_CURSOR is enabled?

heroic-core/src/main/java/com/spotify/heroic/http/metadata/MetadataResource.java
hexedpackets: Yeah, I can add it to that path too. Just needed to have something in there for the updated constructor.

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/AbstractElasticsearchMetadataBackend.java
hexedpackets: I don't like having generics separated like that - makes me double-take to realize its a `BiFunction<Thing>` rather than just a `BiFunction`. AFAIK there's no standard/best-practice style guide for this.

What about splitting in the middle of the generic sig, like:

```
BiFuncton<String, SearchHit,
          Supplier<AsyncFuture<SearchResponse>>> searchFactory
```

metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
hexedpackets: Yep I have the logic backwards, will fix.

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/AbstractElasticsearchMetadataBackend.java
dmichel1: that's more readable for me 👍

heroic-core/src/main/java/com/spotify/heroic/http/metadata/MetadataResource.java
dmichel1: I don't think it matters too much - once we test this on the query metric path, it will likely become the default or only choice.

hexedpackets: Cool - there's no feature support in any `/metadata` queries right now, so its actually kind of annoying to add it in. I'll skip it.

PR #647, opened 2020-04-24
[metrics-bigtable] Run IT tests in the Bigtable emulator.

hexedpackets:


README.md
```diff
@@ -136,46 +136,15 @@ pull request, otherwise it will be rejected by Travis.
 
 #### Remote Integration Tests
 
-Integration tests are configured to run remotely depending on a set of system
+Integration tests are configured to run remotely for Datastax depending on a set of system
```
dmichel1, line 139: I don't think this is true anymore either. [pr 605](https://github.com/spotify/heroic/pull/605/files#diff-230b16a83c49e08a9dedfdd44b97ae46R13) switched it out to test containers

Issue #634, opened 2020-04-17
Partial elasticsearch results should not be returned


dmichel1:
If the find series search completes with errors the request should fail instead of returning partial results. 

The default should be not to return partial results but can be overriden. Perhaps this can be done using the features capability of Heroic.

Related to https://github.com/elastic/elasticsearch/issues/28494

PR #648, opened 2020-04-27
[metadata] do not allow partial results from elasticsearch

dmichel1:
Heroic should not return partial results as it will mislead users and cause false positive alerts.

Now when an elasticsearch shard is completely unavailable (including any replicas) the search will fail.

Heroic will return this failure as part of the response body.

```
      "errors": [
        {
          "type": "shard",
          "nodes": [
            "LocalClusterNode(localMetadata=NodeMetadata(version=0, id=be4b35e7-df70-48a2-a263-1c201cdc9248, tags={site=foo}, service=ServiceInfo(name=The Heroic Time Series Database, version=0.0.1-SNAPSHOT, id=heroic, commit=ac849fac371ee142394f0f48165ce0d43c5dfb74)))"
          ],
          "shard": {
            "site": "guc"
          },
          "error": "[es01][172.21.0.3:9300][indices:data/read/search], caused by [es01][172.21.0.3:9300][indices:data/read/search], caused by Partial shards failure, caused by [es02][172.21.0.2:9300] Node not connected"
        }
      ]
```

The `what: find-series` metric will show the failures as well.

Closes https://github.com/spotify/heroic/issues/634

I can't find a way to get the single node ES cluster that is spun up as part of IT test to go red. Also writing unit tests was too much mocking for the value they added.

### DIFF
--- a/metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
+++ b/metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
@@ -451,7 +451,9 @@ public class MetadataBackendKV extends AbstractElasticsearchMetadataBackend
         OptionalLimit limit = seriesRequest.getLimit();
 
         return doto(c -> {
-            SearchRequest request = c.getIndex().search(METADATA_TYPE);
+            SearchRequest request =
+                c.getIndex().search(METADATA_TYPE).allowPartialSearchResults(false);
+
             request.source()
                 .size(limit.asMaxInteger(scrollSize))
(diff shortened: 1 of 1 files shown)


sjoeboo:
@dmichel1 LGTM. As for getting a 1 node "cluster" to go red...can you increase the replica count (so replica shards should be unassigned...?) or, tell it to expect more masters/more masters for quorum ?

dmichel1:
@sjoeboo I tried increasing the replica count but that makes the cluster yellow. 

how i got this working locally is spinning up two ES nodes, making the replica count 0, shards > 2 and then taking a node offline.

`allowPartialSearchResults(false)` will then cause the search to fail since the missing shards aren't replicated.

PR #650, opened 2020-04-28
Run system tests in a docker full environment.

hexedpackets:
I setup the workflow to run in a nightly job, but while testing I temporarily had an always-run workflow. Here's the output of it: https://circleci.com/workflow-run/fd3d0ff0-fd70-462c-937b-ffdb160db110

I think it makes sense to add this to the normal PR CI. The docker build takes about 6 minutes and could/should replace the Docker Hub build. The docker-compose setup and tests take another ~5 minutes on top of that.

Right now the docker-compose setup runs all possible modules except SRV cluster discovery and memcached for caching. As we add tests we might need to spin up multiple clusters with different configurations, but I think that should be pretty straight forward.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -90,16 +90,11 @@ workflows:
             - build
             - validate_branch
 
-  daily:
-    jobs:
-      - docker_build
+      - docker_build:
+          requires:
+            - validate_branch
+
       - system_test:
           requires:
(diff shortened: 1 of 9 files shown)


hexedpackets:
I realized the system tests can run in parallel with the rest of the build, so I'm going to update the circle config.

PR #661, opened 2020-05-22
[elasticsearch-metadata] Add hash to use for sorting.

### LINKED ISSUE #659: Metadata pagination isn’t returning same results as scrolling
The issue is not consistent, but some queries return more series from Elasticsearch when using scrolls instead of search after pagination.

hexedpackets:
Adds the hash which was already used as `_id` for elasticsearch, but in an indexed field. This also changes the sort field for searching so it will need to be deployed to consumers before API nodes.

Close #659.


sjoeboo:
LGTM!

PR #679, opened 2020-07-20
[Elasticsearch] Fix reporting of dropped by duplicate metrics

lmuhlha:


malish8632:
@lmuhlha just to make sure I didn't miss it. The biggest changes I see you removed try() for tracer.withSpan and rapped connection.doto to private doto. Is there anything else big I've missed?

lmuhlha:
@malish8632 most of the changes are cosmetic when i was comparing the classes. the important change with the fix is 
https://github.com/spotify/heroic/pull/679/files#diff-96ff9d0380bf9df05d420452e2a824eaR57 
and
https://github.com/spotify/heroic/pull/679/files#diff-1290b46b7a9f097ecbf07d898c73e788R681

The problem was that the version conflict format changed from the transport client to the rest client, and on top of that it appears to be nested inside another error. So it was being caught as a failure rather than dropped by duplicate. 

I can revert the cosmetic changes if need be, I just left them because I liked the more similar formatting.

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/AbstractElasticsearchBackend.java
```diff
@@ -56,7 +54,7 @@ public void onFailure(Exception e) {
         Provider<T> emptyProvider, Runnable reportWriteDroppedByDuplicate
     ) {
         return throwable -> {
-            if (ExceptionUtils.getRootCause(throwable) instanceof VersionConflictEngineException) {
+            if (throwable.getMessage().contains("version conflict")) {
```
smstone, line 57: Since we know what exceptions we want to deal with, why not just have a list of exceptions we act on, rather than string matching? I just wonder if the text will change at some point, which would cause this to break.

lmuhlha: Yeah that's what I was thinking as well. I think I tried going into the nested exception but I couldn't the other day.
Let me run it locally again and grab the exact errors and try again.

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/AbstractElasticsearchBackend.java
```diff
@@ -60,6 +60,9 @@ public void onFailure(Exception e) {
                 // Index request rejected, document already exists. That's ok, return success.
```
ao2017, line 60: Same has below you can combine in a single statement

PR #690, opened 2020-09-03
Add total stats across all queries in local manager

malish8632:
This should help understand the behavior of heroic under pressure.
Next step would be to add these stats as Gauges to SemanticRegistry and use them also to reject queries if above global limit.
Currently each query is responsible for exiting in case limits are breached.

### DIFF
--- a/heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
@@ -66,6 +66,7 @@ import io.opencensus.trace.Span;
 import io.opencensus.trace.Tracer;
 import java.util.ArrayList;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Optional;
@@ -104,6 +105,7 @@ public class LocalMetricManager implements MetricManager {
     private final MetricBackendReporter reporter;
     private final QueryLogger queryLogger;
     private final Semaphore concurrentQueries;
+    private static final HashSet<QuotaWatcher> quotaWatchers = new HashSet<>();
 
     /**
      * @param groupLimit The maximum amount of groups this manager will allow to be generated.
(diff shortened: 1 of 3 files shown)


heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -737,7 +747,9 @@ public void readData(long n) {
         public void retainData(final long n) {
             long curRetainedDataPoints = retained.addAndGet(n);
             if (curRetainedDataPoints > 500_000) {
-                log.info("Current retained data points num: {}", curRetainedDataPoints);
+                long total = quotaWatchers.stream().map(QuotaWatcher::getRetainData).reduce(0L, Long::sum);
+                log.info("Total: {}; watchers: {}; Data points retained: {}; added: {}", total,
```
sming, line 751: Can we distinguish the log.info messages a bit more? I have to read most of the line to distinguish between retained and read data. Also, let's spell-out the different numbers E.g.

```log
blah.info: Sep 3, 2020...: Data Points READ: Instance Total: x, This Query Total: y, Delta: z (# Watchers: q)
...
blah.info: Sep 3, 2020...: Data Points RETAINED: Instance Total: x, This Query Total: y, Delta: z (# Watchers: q)
```

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -372,6 +374,9 @@ public QueryTrace buildTrace() {
                 dataInMemoryReporter
             );
 
+            // Add watcher to set to compute stats across all queries
+            quotaWatchers.add(quotaWatcher);
```
sming, line 378: I'm having trouble understanding why this line wasn't there in the first place. Do we see any Quota-y stuff in the logs?

sming (changes requested):
Just a couple of logging format things, otherwise LGTM.

Is there any way we can track and log when the `read` and `retained` numbers go down? That would be equally useful (if not moreso?) information IMO.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
malish8632: I've just introduced this hashset specifically to get global numbers

sming (approved):
typo on 741 & 756, but I'm cool with just merging it.

PR #693, opened 2020-09-08
Moving quota watcher clean up to cover all queries

malish8632:
Moved the code to remove quota watcher at the beginning of end() method to make sure we cover all queries.
Some failures would be keeping watchers in the hash and skewing global stats.


sming (approved):
LGTM

PR #696, opened 2020-09-16
Add global data points stats

malish8632:
This will allow us to observe the data points via Grafana and also via logs.
The premise of this to identify global limits we could use to protect Heroic from OOMing by rejecting requests in case current global data points are above the limits.

### DIFF
--- a/heroic-component/src/main/java/com/spotify/heroic/statistics/MetricBackendReporter.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/statistics/MetricBackendReporter.java
@@ -33,4 +33,8 @@ public interface MetricBackendReporter {
     FutureReporter.Context reportQueryMetrics();
 
     void reportWritesDroppedBySize();
+
+    void reportGlobalReadDataPoints(long points);
+
+    void reportGlobalRetainedDataPoints(long points);
 }
--- a/heroic-component/src/main/java/com/spotify/heroic/statistics/MetricBackendReporter.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/statistics/MetricBackendReporter.java
@@ -34,7 +34,7 @@ public interface MetricBackendReporter {
 
     void reportWritesDroppedBySize();
 
-    void reportGlobalReadDataPoints(long points);
+    void reportTotalReadDataPoints(long points);
 
-    void reportGlobalRetainedDataPoints(long points);
+    void reportTotalRetainedDataPoints(long points);
 }
--- a/heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
@@ -436,11 +436,13 @@ public class LocalMetricManager implements MetricManager {
         private void cleanQuotaWatchers(QuotaWatcher quotaWatcher) {
             // Let's find any old watchers based on start timestamp
             quotaWatchers.remove(quotaWatcher);
-            log.info("Millis QuotaWatcher was alive: {}", (System.currentTimeMillis() - quotaWatcher.startMS));
+            log.info("Millis QuotaWatcher was alive: {}",
+                (System.currentTimeMillis() - quotaWatcher.startMS));
 
             quotaWatchers.removeIf(qw -> {
                 if (System.currentTimeMillis() - qw.startMS > 120_000) {
-                    log.info("Removing QuotaWatcher with MS alive: {}", (System.currentTimeMillis() - qw.startMS));
+                    log.info("Removing QuotaWatcher with MS alive: {}",
+                        (System.currentTimeMillis() - qw.startMS));
                     return true;
                 }
                 return false;
--- a/heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
@@ -62,6 +62,7 @@ import eu.toolchain.async.AsyncFramework;
 import eu.toolchain.async.AsyncFuture;
 import eu.toolchain.async.LazyTransform;
 import eu.toolchain.async.StreamCollector;
+import eu.toolchain.async.FutureDone;
 import io.opencensus.trace.Span;
 import io.opencensus.trace.Tracer;
 import java.util.ArrayList;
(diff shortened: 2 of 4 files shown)


heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -713,32 +745,34 @@ public Histogram getRowDensityHistogram() {
         }
     }
 
-    private static class QuotaWatcher implements FetchQuotaWatcher, RetainQuotaWatcher {
+    private class QuotaWatcher implements FetchQuotaWatcher, RetainQuotaWatcher {
```
sming, line 748: What's the rationale for removing the `static`? Presumably this class is always referenced from within the enclosing class?

heroic-component/src/main/java/com/spotify/heroic/statistics/MetricBackendReporter.java
```diff
@@ -33,4 +33,8 @@
     FutureReporter.Context reportQueryMetrics();
 
     void reportWritesDroppedBySize();
+
+    void reportGlobalReadDataPoints(long points);
```
sming, line 37: Not sure about `Global` as to the uninitiated it sounds like it's all clusters i.e. geographically Global. How about `All` or `Total`?

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -410,9 +411,43 @@ public QueryTrace buildTrace() {
...
+
+                    @Override
+                    public void cancelled() throws Exception {
+                        cleanQuotaWatchers(quotaWatcher);
+                        log.info("onDone.cancelled {}", quotaWatchers.size());
+                    }
+                });
         }
 
+        private void cleanQuotaWatchers(QuotaWatcher quotaWatcher) {
```
sming, line 436: This method breaks the Single Responsibility contract i.e. it performs 2 tasks - removing the supplied watcher *and* removing old watchers. Can we inline line 438? That would work.

heroic-component/src/main/java/com/spotify/heroic/statistics/MetricBackendReporter.java
sming: Also where's the change to a concurrent map? Can't find it.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticMetricBackendReporter.java
```diff
@@ -196,6 +211,16 @@ public void reportWritesDroppedBySize() {
         writesDroppedBySize.inc();
     }
 
+    @Override
+    public void reportGlobalReadDataPoints(long points) {
+        globalReadDataPoints.set(points);
```
sming, line 216: `reportX` to me means to log or print something i.e. To generate a report. Any reason we're not using the traditional `setGlobalReadDataPoints`?

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -811,7 +846,7 @@ boolean isRetainQuotaViolated() {
             return retained.get() >= retainLimit;
         }
 
-        private static int getLeft(long limit, long current) {
+        private int getLeft(long limit, long current) {
```
sming, line 849: Why remove static? It uses no object state, right...?

sming (changes requested):
Just some naming stuff

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
malish8632: I don't think so as we instantiating this class [here](https://github.com/spotify/heroic/blob/master/heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java#L371).

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticMetricBackendReporter.java
malish8632: The naming is inline of SemanticMetricBackendReporter as the object name implies.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
malish8632: This was the reason the whole class was marked as static. There is no obvious reason why it should be static as the methods using this method called on instance rather than static context.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
malish8632: Done

heroic-component/src/main/java/com/spotify/heroic/statistics/MetricBackendReporter.java
malish8632: Names are changed.
I've changed the implementation and using Collection.removeIf method that was suggested to use in this case.

statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticMetricBackendReporter.java
sming: Hmm. OK.

sming (changes requested):
> This was the reason the whole class was marked as static. There is no obvious reason why it should be static as the methods using this method called on instance rather than static context.

Strongly disagree - this has nothing to do with the class being marked `static`. Any non-static class can have any number of static methods. 
Furthermore, it's good practice to mark methods that do not utilise object state as `static` since that makes them easier to unit test.

However it's not a big deal so you can resolve as-is if you like 👍🏻

But... this one : 

> I've changed the implementation and using Collection.removeIf method that was suggested to use in this case.

I do not understand at all. `removeIf` uses an `iterator`, which will be invalidated if the collection is modified under its feet. This doesn't solve the problem, unless I'm missing something...? I see a lot of `Concurrent...` usage in the class but `quotaWatchers` is still just a `HashMap`. Right? : 

```
    private static final HashSet<QuotaWatcher> quotaWatchers = new HashSet<>();
```
?

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -376,7 +374,7 @@ public QueryTrace buildTrace() {
             );
 
             // Add watcher to set to compute stats across all queries
-            quotaWatchers.add(quotaWatcher);
+            quotaWatchers.put(quotaWatcher.hashCode(), quotaWatcher);
```
sming, line 381: So about hashCode... This *should* be fine but the implementation is JVM dependent. It is also **not** simply driven by the the memory address since objects move during GC... 

So, I'd be happier if we'd defined hashCode but it will likely work as-is. Your call @malish8632 🙂

sming (commented):
Observation - your call whether it's ok.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
ao2017: This is a utility  method keep it static

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
ao2017: @malish8632 this doesn't look right to me. In a multi-threaded env,  this will have some consequence on the number that is collected.  It is indeed always reference from this class.   You also need to consider  memory management implication.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
malish8632: It's not possible to keep it static anymore. I've tested this change locally by sending tons of request and I've deployed this as individual container to production. None of the regular metrics - memory or stats were affected. 
We will have to observe for any anomalies but this is the best I can do in regards of verifying this change.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
malish8632: I think the hashCode observation is important but as long as it works for this solution it's fine. Let's deploy and observe. This is pre-requisite for change to act on those stats - having this will help us understand the behavior better.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
malish8632: I've looked already into this with regards to memory - this object is instantiated rather than used as static.
Looking into original implementation the static method getLeft() was the only method marked as static but it was always used to compute on instance variables which to be honest doesn't make sense unless I'm missing something.
And to confirm that I inadvertently remove something important I've tested this code by running it for couple of days on production and can conclude that this change didn't affect much.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
sming: 👍🏻

sming (approved):
LGTM

PR #698, opened 2020-09-23
Add Bigtable profile to metrics and analytics modules

malish8632:
This is required to configure Heroic to use different clusters specifically to separate read and writes after introducing replica.
[App profiles](https://cloud.google.com/bigtable/docs/app-profiles) are used to route clients to different clusters on the same instance. 
Bigtable instances could have multiple clusters.

All tests are successful as well locally verified configuration with and without application `profile` ID.

### DIFF
--- a/metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableConnectionBuilder.java
+++ b/metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableConnectionBuilder.java
@@ -124,7 +124,6 @@ public class BigtableConnectionBuilder implements Callable<BigtableConnection> {
 
     public String toString() {
         return "BigtableConnectionBuilder(project=" + this.project + ", instance=" + this.instance
-            + ", profile=" + (this.profile != null ? this.profile : "null")
-            + ", credentials=" + this.credentials + ")";
+            + ", profile=" + this.profile + ", credentials=" + this.credentials + ")";
     }
 }
--- a/metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableConnectionBuilder.java
+++ b/metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableConnectionBuilder.java
@@ -43,6 +43,7 @@ public class BigtableConnectionBuilder implements Callable<BigtableConnection> {
 
     private final String project;
     private final String instance;
+    private final String profile;
 
     private final CredentialsBuilder credentials;
 
@@ -56,6 +57,7 @@ public class BigtableConnectionBuilder implements Callable<BigtableConnection> {
     public BigtableConnectionBuilder(
         final String project,
         final String instance,
+        final String profile,
         final CredentialsBuilder credentials,
         @Nullable final String emulatorEndpoint,
         final AsyncFramework async,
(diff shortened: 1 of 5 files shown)


lmuhlha:
So this goes from specifying a cluster to specifying an instance. Where does one specify the cluster they wish to use now?

metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableConnectionBuilder.java
```diff
@@ -117,6 +124,7 @@ public BigtableConnection call() throws Exception {
 
     public String toString() {
         return "BigtableConnectionBuilder(project=" + this.project + ", instance=" + this.instance
-               + ", credentials=" + this.credentials + ")";
+            + ", profile=" + (this.profile != null ? this.profile : "null")
```
ao2017, line 127: I think this is unnecessary => "profile =" + this.profile  is sufficient.

metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableMetricModule.java
```diff
@@ -126,7 +129,7 @@ public Exposed module(PrimaryComponent primary, Depends depends, String id) {
                 public AsyncFuture<BigtableConnection> construct() {
                     return async.call(
                         new BigtableConnectionBuilder(
-                            project, instance, credentials, emulatorEndpoint,
+                            project, instance, profile, credentials, emulatorEndpoint,
```
ao2017, line 132: Is "emulatorEndpoint" for testing ?

malish8632: yeah, that was implemented by William

metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableConnectionBuilder.java
malish8632: I wasn't sure if null pointer exception will kick in here if it is null.

malish8632: fixed

metric/bigtable/src/main/java/com/spotify/heroic/metric/bigtable/BigtableMetricModule.java
```diff
@@ -87,6 +89,7 @@ public BigtableMetricModule(
         this.groups = groups.orElseGet(Groups::empty).or(DEFAULT_GROUP);
         this.project = project.orElseThrow(() -> new NullPointerException("project"));
         this.instance = instance.orElse(DEFAULT_INSTANCE);
+        this.profile = profile.orElse(null);
```
lmuhlha, line 92: should we have a default profile here instead of null?

malish8632: I thought about this but ended up using default without literally setting it up to "default". In BigtableConnectionBuilder it would behave the same way as before in case it is not set in config.

lmuhlha (approved):

ao2017 (approved):

PR #699, opened 2020-10-01
[WIP]Add distribution support to spotify100_proto and BigtableBackend writer

ao2017:
This is the first of several PRs  that will add distribution support to heroic.  With Distribution support heroic will be able to provide histogram data based on data distribution. Distribution support will also enable Opencensus metric import. 
This PR  implements Spotify100Proto consumer and BigtableBackend(ingestion).
**What is New:**
1. Spotify100 Proto Consumer:
  This PR introduces a new dataPoint type.
  ```
@AutoValue
public abstract class DistributionPoint implements Metric {

    public abstract long getTimestamp();
    public abstract Distribution value();

    public static DistributionPoint create(final Distribution value, final long timestamp) {
        return new AutoValue_DistributionPoint(timestamp, value);
    }

    @Override
    public boolean valid() {
        return value().getValue() != null &&  !value().getValue().isEmpty();
    }

    @Override
    public void hash(@NotNull Hasher hasher) {
        hasher.putInt(this.hashCode());
    }
} 
```
2. BigTableBackend
This PR introduces a mechanism to insert DistributionPoint into Bigtable. 
The overall structure of the BigTableBackend does not change. 
But this PR introduces a new column family "DistributionPoint"

Referecence:
https://github.com/spotify/heroic/issues/695

### DIFF
--- a/heroic-core/src/main/proto/spotify_100.proto
+++ b/heroic-core/src/main/proto/spotify_100.proto
@@ -20,8 +20,24 @@ message Metric {
 
   // resource "tags" associated to metric.
   map<string, string> resource = 5;
+
+  // Distribution type value
+  Value distribution_type_value = 6;
+}
+
+message Message {
+  Metric metric = 1;
 }
 
 message Batch {
   repeated Metric metric = 1;
 }
+
+
+message Value {
+  oneof value {
+    double double_value = 1;
+    bytes distribution_value= 2;
+  }
+}
+
--- a/heroic-core/src/main/proto/spotify_100.proto
+++ b/heroic-core/src/main/proto/spotify_100.proto
@@ -25,10 +25,6 @@ message Metric {
   Value distribution_type_value = 6;
 }
 
-message Message {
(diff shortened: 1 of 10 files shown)

### CODE REVIEW

File: heroic-core/src/main/proto/spotify_100.proto [hunk 1]
```diff
@@ -20,8 +20,24 @@ message Metric {
 
   // resource "tags" associated to metric.
   map<string, string> resource = 5;
+
+  // Distribution type value
+  Value distribution_type_value = 6;
+}
+
+message Message {
```
malish8632, line 28: I just want to point out - this Message object is never used in either FFWD or here. 
1. Maybe we shouldn't add this here as it is not used?
2. And maybe we should remove it from FFWD proto definition to make it identical?
Thoughts?

ao2017: Done

malish8632 (approved):
LGTM

Issue #695, opened 2020-09-15
Add Distribution Support to Heroic Consumer


ao2017:
Heroic histogram data is currently computed locally.
It is practically impossible to aggregate percentile.
We are adding distribution to heroic to address that issue.  Applications downstream will create data sketches  that can be merged to compute percentile on the entire data distribution.
To support this approach, we  changed the format of the metric. 
Heroic consumers should be able to handle both the old and the new metrics.  
We are creating a new version of Spotify100 serializer. This task will implement the logic to handle spotify100 new version metrics.
**Reference**
Current Format
Spotify100 version SCHEMA_VERSION = "1.1.0"
 @Data
    public static class Spotify100Metric {
        private final String version = SCHEMA_VERSION;
        private final String key;
        private final Long time;
        private final Map<String, String> attributes;
        private final Map<String, String> resource;
        private final double  value;
    }
Expected Format
Spotify100 version SCHEMA_VERSION = "1.1.2" (actual TBD)
 @Data
    public static class Spotify100Metric {
        private final String version = SCHEMA_VERSION;
        private final String key;
        private final Long time;
        private final Map<String, String> attributes;
        private final Map<String, String> resource;
        private final Value  value;
    }

**Code Pointer:**
https://github.com/spotify/heroic/blob/5af661539d8a2f546433f95556aa5724bda8bf73/heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100.java#L137

PR #700, opened 2020-10-07
[WIP] Add distribution support to spotify100 and the ability to fetch distribution from bigTable. 

ao2017:
This PR add distribution support to spotify100 (jsonMetric) and the ability to fetch distribution metric from bigTable.
Spotify100 consumers and ingestion modules should now be able to handle new and old JsonMetric.
For more information please refer to this [issue](https://github.com/spotify/heroic/issues/695)

### DIFF
--- a/heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100.java
@@ -23,6 +23,7 @@ package com.spotify.heroic.consumer.schemas;
 
 import static io.opencensus.trace.AttributeValue.stringAttributeValue;
 
+
 import com.fasterxml.jackson.databind.JsonNode;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.databind.node.JsonNodeType;
@@ -38,9 +39,12 @@ import com.spotify.heroic.consumer.ConsumerSchemaValidationException;
 import com.spotify.heroic.consumer.SchemaScope;
 import com.spotify.heroic.consumer.schemas.spotify100.JsonMetric;
 import com.spotify.heroic.consumer.schemas.spotify100.Version;
+import com.spotify.heroic.consumer.schemas.spotify100.v2.Value;
 import com.spotify.heroic.ingestion.Ingestion;
 import com.spotify.heroic.ingestion.IngestionGroup;
 import com.spotify.heroic.ingestion.Request;
+import com.spotify.heroic.metric.DistributionPoint;
+import com.spotify.heroic.metric.HeroicDistribution;
 import com.spotify.heroic.metric.MetricCollection;
 import com.spotify.heroic.metric.Point;
 import com.spotify.heroic.statistics.ConsumerReporter;
@@ -126,6 +130,8 @@ public class Spotify100 implements ConsumerSchema {
 
                 if (version.getMajor() == 1) {
                     return handleVersion1(tree).onFinished(span::end);
+                } else if (version.getMajor() == 2) {
+                    return handleVersion2(tree).onFinished(span::end);
                 }
 
                 span.setStatus(Status.INVALID_ARGUMENT.withDescription("Unsupported version"));
@@ -178,6 +184,76 @@ public class Spotify100 implements ConsumerSchema {
             reporter.reportMetricsIn(1);
             AsyncFuture<Ingestion> ingestionFuture =
                 ingestion.write(new Request(series, MetricCollection.points(points)));
+
+            // Return Void future, to not leak unnecessary information from the backend but just
+            // allow monitoring of when the consumption is done.
+            return ingestionFuture.directTransform(future -> null);
+        }
+
+        private AsyncFuture<Void> handleVersion2(final JsonNode tree)
+            throws ConsumerSchemaValidationException {
+            final com.spotify.heroic.consumer.schemas.spotify100.v2.JsonMetric metric;
+
+            try {
+                metric = new TreeTraversingParser(tree, mapper)
+                    .readValueAs(
+                        com.spotify.heroic.consumer.schemas.spotify100.v2.JsonMetric.class);
+            } catch (IOException e) {
+                throw new ConsumerSchemaValidationException("Invalid metric", e);
+            }
+
+            if (metric.getValue() == null) {
+                throw new ConsumerSchemaValidationException(
+                    "Metric must have a value but this metric has a null value: " + metric);
+            }
+
+            if (metric.getTime() == null) {
+                throw new ConsumerSchemaValidationException("time: field must be defined: " + tree);
+            }
+
+            if (metric.getTime() <= 0) {
+                throw new ConsumerSchemaValidationException(
+                    "time: field must be a positive number: " + tree);
+            }
+
+            if (metric.getKey() == null) {
+                throw new ConsumerSchemaValidationException("key: field must be defined: " + tree);
+            }
+
+            final Map<String, String> tags = new HashMap<>(metric.getAttributes());
+
+            if (metric.getHost() != null) {
+                tags.put(HOST_TAG, metric.getHost());
+            }
+
+            final Map<String, String> resource = new HashMap<>(metric.getResource());
+
+            final Series series = Series.of(metric.getKey(), tags, resource);
+
+            AsyncFuture<Ingestion> ingestionFuture;
+            Value value = metric.getValue();
+            if (value instanceof Value.DoubleValue) {
+                final Value.DoubleValue doubleValue = (Value.DoubleValue) value;
+                final double val = doubleValue.getValue();
+                final Point p = new Point(metric.getTime(), val);
+                final List<Point> points = ImmutableList.of(p);
+                ingestionFuture =
+                    ingestion.write(new Request(series, MetricCollection.points(points)));
+            } else if (value instanceof Value.DistributionValue) {
+                final Value.DistributionValue distributionValue = (Value.DistributionValue) value;
+                final HeroicDistribution heroicDistribution =
+                    HeroicDistribution.create(distributionValue.getValue());
+                final DistributionPoint point = DistributionPoint.create(heroicDistribution,
+                    metric.getTime());
+                ingestionFuture = ingestion.write(new Request(series,
+                    MetricCollection.distributionPoints(List.of(point))));
+            } else {
+                throw new RuntimeException("Unknown type value = " + value);
+            }
+
+            reporter.reportMessageDrift(clock.currentTimeMillis() - metric.getTime());
+            reporter.reportMetricsIn(1);
+
 
             // Return Void future, to not leak unnecessary information from the backend but just
             // allow monitoring of when the consumption is done.
--- a/heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100.java
@@ -128,9 +128,11 @@ public class Spotify100 implements ConsumerSchema {
                     throw new ConsumerSchemaValidationException("Bad version: " + versionNode);
                 }
 
-                if (version.getMajor() == 1) {
+                int major = version.getMajor();
+
+                if (major == 1) {
                     return handleVersion1(tree).onFinished(span::end);
-                } else if (version.getMajor() == 2) {
+                } else if (major == 2) {
                     return handleVersion2(tree).onFinished(span::end);
                 }
 
--- a/heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/JsonMetric.kt
+++ b/heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/JsonMetric.kt
@@ -1,0 +1,26 @@
+package com.spotify.heroic.consumer.schemas.spotify100.v2
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.annotation.JsonProperty
+
+@JsonIgnoreProperties(ignoreUnknown = true)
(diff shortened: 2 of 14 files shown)


heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100.java
```diff
@@ -179,6 +185,76 @@ public Consumer(Clock clock, IngestionGroup ingestion, ConsumerReporter reporter
...
+            if (metric.getTime() == null) {
+                throw new ConsumerSchemaValidationException("time: field must be defined: " + tree);
+            }
+
+            if (metric.getTime() <= 0) {
+                throw new ConsumerSchemaValidationException(
+                    "time: field must be a positive number: " + tree);
+            }
+
+            if (metric.getKey() == null) {
```
sming, line 219: 70% of `handleVersion2()` is copy pasta from `handleVersion1()`. Please either explain why this is, or refactor the common 70%.

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/Value.java
```diff
@@ -0,0 +1,69 @@
...
+import com.google.auto.value.AutoValue;
+import com.google.protobuf.ByteString;
+
+/**
+ * Distribution histogram point value. Currently we
+ * support {@link Value.DoubleValue} and {@link Value.DistributionValue}
+ */
+@JsonSerialize(using = ValueSerializer.class)
+@JsonDeserialize(using = ValueDeserializer.class)
+public  abstract class Value {
```
sming, line 38: typo - double space

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/JsonMetric.kt
```diff
@@ -0,0 +1,26 @@
+package com.spotify.heroic.consumer.schemas.spotify100.v2
+
+import com.fasterxml.jackson.annotation.JsonIgnoreProperties
+import com.fasterxml.jackson.annotation.JsonProperty
+
+@JsonIgnoreProperties(ignoreUnknown = true)
+data class JsonMetric(
```
sming, line 7: please leave a class comment - e.g. what it's for and where it's used.

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/Value.java
sming, line 38: please leave a class comment - e.g. what it's for and where it's used.

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/Value.java
```diff
@@ -0,0 +1,69 @@
...
+        @JsonCreator
+        public static Value.DistributionValue create(
+            final ByteString value) {
+            return new AutoValue_Value_DistributionValue(value);
+        }
+
+        @Override
+        public abstract ByteString getValue();
+    }
+
```
sming, line 68: remove blank line (extra)

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/ValueDeserializer.java
```diff
@@ -0,0 +1,69 @@
...
+    @Override
+    public Value deserialize(JsonParser jsonParser, DeserializationContext deserializationContext)
+        throws IOException, JsonProcessingException {
+
+        JsonNode node = jsonParser.getCodec().readTree(jsonParser);
+
+        if (node.get("distributionValue") != null) {
+            byte[] bytes = node.get("distributionValue").binaryValue();
+            ByteString byteString = ByteString.copyFrom(bytes);
+            return Value.DistributionValue.create(byteString);
```
sming, line 59: if you combine these calls into 1 line, the JVM JIT will be able to perform [return value optimization](https://stackoverflow.com/a/42047862/1170932).

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/ValueSerializer.java
```diff
@@ -0,0 +1,69 @@
...
+public class ValueSerializer extends StdSerializer<Value> {
+
+    private static final long serialVersionUID = 6300597228325654588L;
+
+    public ValueSerializer() {
+        this(Value.class);
+    }
+
+
+
```
sming, line 43: please tidy up. Run `checkstyle` on your modifications 👍🏻

sming (commented):
LGTM. I think there is perhaps a missed opportunity to use [polymorphism instead of conditional](https://www.refactoring.com/catalog/replaceConditionalWithPolymorphism.html) i.e. replace all the : 
```java
Foo foo = ...;

if (foo.getVersion() == 1) {
    ...
} else if (foo.getVersion() == 2) {
    ...
}
```

with subclasses of foo's type:
```java
 Foo
  ↑
FooV1 FooV2
```

it'd be a bit of work to implement but it would reap a lot of benefits. @ao2017 thoughts?

malish8632:
Hey @ao2017 I think you changed [Spotify100.java](https://github.com/spotify/heroic/blob/master/heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100.java) schema but this is old schema and not used atm. The main schema that is used now is [Spotify100Proto.java](https://github.com/spotify/heroic/blob/master/heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100Proto.java)

ao2017:
We already committed to doing http and protoBuf clients so let's move forward with the PR.
 From what I know, ffwd-http and ffwd-json modules  send metric to Heroic in Spotify100 format. 
There is no side effect in having distribution support in Spotify100.

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/ValueSerializer.java
ao2017: @sming if I am not mistaken, checkstyle is part of the build process.

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100.java
ao2017: Agreed, I thought about it but decided not to change anything. it is too much work for little return.
We need to eventually take on the task to streamline Metric definition may be refactoring can be done with that task.

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/ValueDeserializer.java
ao2017: @sming which optimization. Isn't this more readable :). The compiler will probably re-arrange this. Anyway done!

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/Value.java
ao2017: Please kindly ignore, I will probably remove this eventually. The way the code is structure at the moment metric is defined at every stage. I think  we will eventually use the definition in semantic-metric it has all the  documentation.

malish8632:
@ao2017 Spotify100 schema is not used on Heroic side at all anymore. FFWD does support JSON but it internally converts that to spotify_100.proto when publishes it to PubSub. The changes you've added are never going to be used in Heroic and I don't see the reason to keep that code.

ao2017:
Sergey, Is this dead code ?https://github.com/spotify/ffwd/blob/b5c9837a06abab68e1322ae0848090403a345c9a/modules/http/src/main/java/com/spotify/ffwd/http/HttpClient.java#L59

malish8632:
@ao2017 this is not a dead code. This code would be used by locally configured FFWD agent to sent metrics via HTTP/JSON to metrics-api, which in itself a FFWD instance configured to get HTTP requests and convert them to spotify_100.proto to be published to PubSub. So that code won't send it directly to Heroic. Heroic consumers are the only instances responsible for consumption of spotify_100.proto schema which in itself never will use Spotify100.java code.

ao2017:
Got it, can you point me to the part where the conversion is done.

Anyway, If Spotify100 is not used at all then we should eventually do some clean up and re-write consumers integration tests. Consumers integration tests used spotify100.  For now, I say let's move forward with the PR because it is not just spotify100 consumer.

malish8632:
I gave you some pointers offline w.r.t. code location. 
Don't you need to update Spotify100Proto.java to support distributions?

I'm all for cleaning the code at the same time as it is open source project we would need to go through deprecation routine which we probably should.

ao2017:
Spotify100Proto is done. Please let's move forward with the PR.

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100.java
sming: I just don't agree, to be honest.

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/Value.java
sming: Can you put that as a comment then? It's helpful knowledge.

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/ValueDeserializer.java
sming: The optimization I linked to

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/ValueSerializer.java
sming: If it is, then there's a bug as it's not being applied. Can you run it locally using the local checkstyle.xml

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/Value.java
sming: That's useful info, please put in the comment

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100.java
sming: Sorry but I just don't agree. Copy pasted code needs a very strong argument for keeping it.

sming:
@ao2017 I responded to some comments

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/Value.java
ao2017: Can we please not dwell on the small stuffs. This is a  WIP in several repos with several PRs. We need to get the ball moving. As I said I will be replacing this file.

heroic-dist/src/test/java/com/spotify/heroic/PubSubConsumerIT.java
```diff
@@ -19,10 +19,10 @@
  * under the License.
  */
 
-package com.spotify.heroic;
+    package com.spotify.heroic;
```
malish8632, line 22: Space needs to be removed

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/Value.java
sming: I apologise for the accidental double-posts in several threads like this one. My browser crashed and didn't display the first set of posts so I did them again. I can see how they could be perceived as being irritating.

AFAIK, if the PR author doesn't agree with a comment, that's totally fine. They should simply leave an explanatory comment (like [the above](https://github.com/spotify/heroic/pull/700#discussion_r502425289_) and click "Resolve conversation".

malish8632:
Got it @ao2017. I didn't realize the Spotify100Proto was already addressed in other PR. Thanks for clarifications.

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/ValueSerializer.java
ao2017: Yeah, I didn't disable checkstyle check locally.

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/ValueDeserializer.java
ao2017: I read it. How can the the compiler even optimize this statement ? it may re-arrange it, but I don't see why optimization is needed here

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/spotify100/v2/Value.java
ao2017: Apology accepted.

heroic-core/src/main/java/com/spotify/heroic/consumer/schemas/Spotify100.java
ao2017: Alright, I hope you don't feel like the entire PR was a copy and past :)

ao2017:
@malish8632  NP,  we have been through this  with every PR I posted except one. May be reading the issue/task related to the PR or asking questions may help . Let's figure out how to make the process smoother next time.

malish8632 (approved):
I think it looks good to me.

PR #702, opened 2020-10-19
Debug QuotaWatchers creation and cleaning

malish8632:
This change moving creation of QuotaWatcher closer to its usage minimizing chance of exceptional exit before it is used.
Wraps main call to try/catch in order to properly clean QuotaWatcher map.
It also uses QuotaWatcher as key in ConcurrentMap.


sming (approved):
LGTM

PR #718, opened 2020-11-02
GitHub action workflow to build PRs in Docker images.

### LINKED ISSUE #716: Auto-build a Docker image for PR's
### Auto-build a Docker image for PRs

- I am a /monitoring dev
- Who wants to test out his/her PR in staging, with as little manual grunt work as possible
- So that I can be more efficient, innovative, wonderful and happier

#### Design & Implementation Notes

* it would be great for testing in staging before merging to `master`, to catch issues that only surface once deployed
* extremely useful for disruptive/larger PR's e.g. distribution support
* care should be taken to align this infra feature with engineering's offerings/support. We do not want to swim upstream.

hexedpackets:
I did this as GitHub actions instead of CircleCI because of how forks are handled. We want tests to run for every PR, regardless of source. We only want these Docker images to be automatically created for PRs originating from this repo, and not for forks. Doing that in CircleCI might be possible, but this way is simpler and has the benefit of building the image in parallel with the tests running in Circle.

Assuming this works well, we might want to also move our Docker Hub build actions to GitHub actions to reduce the number of places that CI is happening.

Closes #716.

### DIFF
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -26,5 +26,6 @@ jobs:
           password: ${{ secrets.GCR_JSON_KEY }}
       - uses: docker/build-push-action@v2
         with:
+          context: .
           push: true
           tags: ${{ steps.docker_meta.outputs.tags }}
--- a/.github/workflows/docker-build.yml
+++ b/.github/workflows/docker-build.yml
@@ -1,0 +1,29 @@
+name: Docker build workflow
+# NB: This will only run on PRs from the main repo and not forks
+# https://docs.github.com/en/free-pro-team@latest/actions/reference/events-that-trigger-workflows#pull-request-events-for-forked-repositories
+on:
+  pull_request:
+    branches:
+      - master
+jobs:
+  build-docker:
+    name: Build and push Docker image to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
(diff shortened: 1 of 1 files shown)


lmuhlha (approved):

.github/workflows/docker-build.yml
```diff
@@ -0,0 +1,31 @@
...
+    name: Build and push Docker image to Docker Hub
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      # NB: For PRs, this will create two Docker tags: sha-${GIT_SHA} and pr-${PR_NUMBER}
+      # https://github.com/crazy-max/ghaction-docker-meta#overview
+      - id: docker_meta
+        uses: crazy-max/ghaction-docker-meta@v1
+        with:
+          images: gcr.io/xpn-heroic-1/heroic
```
adsail, line 19: Given this is a public repo, should images for PRs push to internal GCR, or Docker Hub?

If i'm understanding Docker Meta properly, this prepares for a push to internal GCR.

hexedpackets: Right, it pushes to the GCR repo which is internal only. I set it up for GCR because I think the PR builds aren't useful for general consumption, and I wanted to avoid the new Docker Hub pull limits. But its a weak opinion, happy to switch it if you disagree.

adsail: I agree with the decision since the use case is for internal testing, and we'd have to take an additional step to then push to GCR which makes the process unnecessarily complicated. Not to mention, PR builds could be broken/faulty, so best to not publicize them.

PR #725, opened 2020-11-12
Optionally write metadata resource identifiers as tags

adsail:
Introduce a config variable to write metadata resource identifiers to elasticsearch in addition to tags. By default, RIs will continue to not be written as tags. If configured as `true`, RIs are appended to the SortedMap of tags and written as tags to elasticsearch.

### DIFF
--- a/metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
+++ b/metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
@@ -129,7 +129,6 @@ public class MetadataBackendKV extends AbstractElasticsearchMetadataBackend
     private static final String TAGS = "tags";
     private static final String TAG_KEYS = "tag_keys";
     private static final String RESOURCE = "resource";
-    private static final String RESOURCE_KEYS = "resource_keys";
     private static final String HASH_FIELD = "hash";
     private static final Character TAG_DELIMITER = '\0';
 
@@ -613,12 +612,6 @@ public class MetadataBackendKV extends AbstractElasticsearchMetadataBackend
           b.startArray(RESOURCE);
           for (final Map.Entry<String, String> entry : series.getResource().entrySet()) {
             b.value(entry.getKey() + TAG_DELIMITER + entry.getValue());
-          }
-          b.endArray();
-
-          b.startArray(RESOURCE_KEYS);
-          for (final Map.Entry<String, String> entry : series.getResource().entrySet()) {
-            b.value(entry.getKey());
           }
           b.endArray();
         }
--- a/metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
+++ b/metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
@@ -90,6 +90,7 @@ import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.SortedMap;
 import java.util.concurrent.Callable;
 import java.util.concurrent.ConcurrentLinkedQueue;
 import java.util.function.Consumer;
@@ -128,6 +129,8 @@ public class MetadataBackendKV extends AbstractElasticsearchMetadataBackend
     private static final String KEY = "key";
     private static final String TAGS = "tags";
     private static final String TAG_KEYS = "tag_keys";
+    private static final String RESOURCE = "resource";
+    private static final String RESOURCE_KEYS = "resource_keys";
     private static final String HASH_FIELD = "hash";
     private static final Character TAG_DELIMITER = '\0';
 
@@ -212,13 +215,19 @@ public class MetadataBackendKV extends AbstractElasticsearchMetadataBackend
             final Scope rootScope = tracer.withSpan(rootSpan);
 
             final Series series = request.getSeries();
-            if (indexResourceIdentifiers) {
-                rootSpan.putAttribute(
-                    "resource_id_count",
-                    AttributeValue.doubleAttributeValue(series.getResource().size())
-                );
-            }
-            
+//            if (indexResourceIdentifiers) {
+//                SortedMap<String, String> tagsAndResource = request.getSeries().getTags();
+//                tagsAndResource.putAll(request.getSeries().getResource());
+//                series = new Series(request.getSeries().getKey(), tagsAndResource, request.getSeries().getResource());
+//
+//                rootSpan.putAttribute(
+//                    "resource_id_count",
+//                    AttributeValue.doubleAttributeValue(series.getResource().size())
+//                );
+//            } else {
+//                series = request.getSeries();
+//            }
+
             final String id = series.hash();
             rootSpan.putAttribute("id", AttributeValue.stringAttributeValue(id));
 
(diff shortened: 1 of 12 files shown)


metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
```diff
@@ -209,7 +211,23 @@ public Groups groups() {
...
+            final Series series;
+            if (indexResourceIdentifiers) {
+                Series initialSeries = request.getSeries();
+                final double numberOfResourceIds = initialSeries.getResource().size();
+
+                rootSpan.putAttribute(
+                    "resource_id_count",
+                    AttributeValue.doubleAttributeValue(numberOfResourceIds)
+                );
+
```
adsail, line 223: @hexedpackets Does the above (L219-222) afford adequate observability?

metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
```diff
@@ -209,7 +211,23 @@ public Groups groups() {
                 .startSpan();
             final Scope rootScope = tracer.withSpan(rootSpan);
 
-            final Series series = request.getSeries();
+            final Series series;
```
malish8632: As discussed offline we could optimize it a bit and add code to `buildContext` method to add resources to ES request.
Something like this

```java
private static void buildContext(final XContentBuilder b, Series series, boolean indexResourceIdentifiers) throws IOException {
        b.field(KEY, series.getKey());

        b.startArray(TAGS);
        for (final Map.Entry<String, String> entry : series.getTags().entrySet()) {
            b.value(entry.getKey() + TAG_DELIMITER + entry.getValue());
        }

        if(indexResourceIdentifiers) {
           for (final Map.Entry<String, String> entry : series.getResource().entrySet()) {
               b.value(entry.getKey() + TAG_DELIMITER + entry.getValue());
           }
        }

        b.endArray();

        b.startArray(TAG_KEYS);
        for (final Map.Entry<String, String> entry : series.getTags().entrySet()) {
            b.value(entry.getKey());
        }
        b.endArray();

        b.field(HASH_FIELD, series.hash());
    }

File: metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java [hunk 3]
```diff
@@ -577,19 +586,29 @@ public DeleteSeries end(
...
+                b.value(entry.getKey() + TAG_DELIMITER + entry.getValue());
+            }
+        }
         b.endArray();
 
         b.startArray(TAG_KEYS);
         for (final Map.Entry<String, String> entry : series.getTags().entrySet()) {
             b.value(entry.getKey());
         }
+        if(indexResourceIdentifiers) {
```
malish8632, line 607: This block just adds the `tags`, `TAG_DELIMITER + entry.getValue()` should be removed.

metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
```diff
@@ -592,6 +610,20 @@ private static void buildContext(final XContentBuilder b, Series series) throws
...
         b.endArray();
 
+        if (indexResourceIdentifiers) {
+          b.startArray(RESOURCE);
+          for (final Map.Entry<String, String> entry : series.getResource().entrySet()) {
+            b.value(entry.getKey() + TAG_DELIMITER + entry.getValue());
+          }
+          b.endArray();
+
+          b.startArray(RESOURCE_KEYS);
```
malish8632, line 620: Do we need this? Can we maybe verify with unit tests? If it's not used we could drop it as it will be additional size in ES.

adsail: I this in reference to the `RESOURCE_KEYS`?

malish8632: yes, sorry I wasn't clear

metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
```diff
@@ -128,6 +128,8 @@
     private static final String KEY = "key";
     private static final String TAGS = "tags";
     private static final String TAG_KEYS = "tag_keys";
+    private static final String RESOURCE = "resource";
+    private static final String RESOURCE_KEYS = "resource_keys";
```
malish8632, line 132: We don't need this one anymore

malish8632 (approved):
small comment, otherwise +1

metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
adsail: ah yes good catch!

metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
```diff
@@ -439,7 +443,15 @@ protected Series toSeries(SearchHit hit) {
         final Iterator<Map.Entry<String, String>> tags =
             ((List<String>) source.get(TAGS)).stream().map(this::buildTag).iterator();
 
-        return Series.of(key, tags);
+        if (indexResourceIdentifiers) {
+            @SuppressWarnings("unchecked")
```
sming, line 447: is the suppression unavoidable?

sming: oh I see it's done above too. I guess just leave it.

metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
```diff
@@ -592,6 +608,14 @@ private static void buildContext(final XContentBuilder b, Series series) throws
...
         b.endArray();
 
+        if (indexResourceIdentifiers) {
+          b.startArray(RESOURCE);
+          for (final Map.Entry<String, String> entry : series.getResource().entrySet()) {
+            b.value(entry.getKey() + TAG_DELIMITER + entry.getValue());
+          }
+          b.endArray();
+        }
+
```
sming, line 618: I feel that the two for loops (599 & 612) can be replaced by the same function (i.e. refactor them). AFAICT They're all the same types involved. So there'd be 2 calls to the function, which would look like :

```java
private static appendData(final XContentBuilder b, Map<String, String> data, String name) {
        b.startArray(name);
        for (var entry : data.entrySet()) {
            b.value(entry.getKey() + name + entry.getValue());
        }
        b.endArray();
}
```
then L599 would become:
`appendData(b, series.getTags(), TAGS)`

Make sense?

sming (commented):
LGTM

metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
lmuhlha: +1 to this suggestion, the code is fairly similar so would be nice to make it more generic if possible

lmuhlha:
👍  nice job. would be good to make that one bit of code more generic so it can be called to in a tag or resource situation, as peter suggested. otherwise LGTM

metadata/elasticsearch/src/main/java/com/spotify/heroic/metadata/elasticsearch/MetadataBackendKV.java
```diff
@@ -577,24 +589,38 @@ public DeleteSeries end(
...
+    private static void appendData(
+      final XContentBuilder b, Map<String, String> data, String name, Character delimiter)
+      throws IOException {
+        b.startArray(name);
+        for (var entry : data.entrySet()) {
+            b.value(entry.getKey() + delimiter + entry.getValue());
+        }
+        b.endArray();
+    }
+
```
adsail, line 623: @lmuhlha @sming Look good?

PR #732, opened 2020-12-18
Implement new Bigtable timeout & retry settings

sming:
#### Use Case Resolved: Heroic pipeline instability and poor user experience for Grafana

- I am a Heroic developer
- who wants to armour up Heroic so that we get fewer alerts
- so that we can focus on features

#### Plan

* look at the [spotify-bigtable-client](https://ghe.spotify.net/bases/bigtable-client-wrapper) repo for "inspiration"
* don't implement a whole new layer of configuration. Focus on the 2 or 3 settings we're immediately interested in.

#### Notes for Reviewer

- only read requests are affected (i.e. Heroic API but not metric Ingestion)
- a maxMutateRowTimeout *is* implemented but will be deployed with the default setting value so as not to change Ingestion at all. This is done for development efficiency reasons i.e. it would have been foolish not to add support for it whilst I was changing the exact same files in the exact same places. Here is the [heroic-gke](https://ghe.spotify.net/monitoring/heroic-gke/pull/111) PR that puts `mutateRpcTimeoutMs` back to 10 minutes.

### DIFF
--- a/docs/content/_docs/config.md
+++ b/docs/content/_docs/config.md
@@ -132,6 +132,7 @@ Precedence for each flag is defined as the following:
 The following features are available:
 
 #### com.spotify.heroic.deterministic_aggregations
+
 {:.no_toc}
 
 Enable feature to only perform aggregations that can be performed with limited resources. Disabled by default.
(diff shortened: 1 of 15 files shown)


sming:
was temporary PR - deleting

PR #734, opened 2020-12-21
Moves Tdigest stat computation to SharedGroups combine phase. (Part 1 of 2)

ao2017:
Can I get a quick review for this PR the actual change and the integration tests will follow.
This is mostly boilerplate changes in Bucket and AggregationSession implementations. 
The PR also add a new metric TdigestPoint, it is similar to group metric. It is used during the last phase of aggregation to 
merge tdigest from different cluster.


lmuhlha:
@ao2017 this looks alright to me, was just wondering since the quantiles are removed in the PR: is that to make the tdigest aggregation more generic and then you'll introduce default quantiles later? 
And why does `TdigestStatInstanceUtils` and `testTDigestStatInstance` continue to have stat in the name? What's the significance of removing it everywhere except there?

ao2017:
> @ao2017 this looks alright to me, was just wondering since the quantiles are removed in the PR: is that to make the tdigest aggregation more generic and then you'll introduce default quantiles later?
> And why does `TdigestStatInstanceUtils` and `testTDigestStatInstance` continue to have stat in the name? What's the significance of removing it everywhere except there?

I moved the quantile to a different file since the computation is moving at a different level. That change will be included in the second PR. I am trying to keep the file count in this PR down. 

I also changed the name of the Utils and class and the test.

lmuhlha (approved):

PR #746, opened 2021-01-22
Dynamic Metadata Read Indices 

### LINKED ISSUE #745: Dynamically determine metadata read indices based upon query range
While the rotation interval of the ES write index—in other words, how frequently Heroic consumers stop writing to an elasticsearch index, create a new one and start writing to it instead—is configurable, the maximum number of read indices is not configurable and not dynamic. 

`maxReadIndices` should be dynamically determined based upon the date range of a query and the rotation interval of Elasticsearch indices. We can then narrow the metadata response to only return documents relevant to the query as opposed to all documents. This will narrow the query to BigTable and lay the groundwork for future guardrails to enable malleable series limits that take query range into consideration.

adsail:
Introduce support for RotatingIndexMapping to limit the number of elasticsearch metadata indices read based upon the range of a heroic query. If the newly introduced config flag is set to `true`, the number of read indices is dynamically determined based upon the rotation interval and the range of the query.

This PR also makes the non-dynamic `maxReadIndices` configurable so as to properly support `delete`. This variable should be set to the maximum number of queryable indices that could possibly exist in Elasticsearch based upon the rotation interval and ILM policy. 

Closes #745.

### DIFF
--- a/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/index/RotatingIndexMapping.kt
+++ b/heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/index/RotatingIndexMapping.kt
@@ -27,19 +27,21 @@ import org.elasticsearch.action.delete.DeleteRequest
 import org.elasticsearch.action.search.SearchRequest
 import org.elasticsearch.action.support.IndicesOptions
 import org.elasticsearch.search.builder.SearchSourceBuilder
-//import org.slf4j.LoggerFactory
+import org.slf4j.LoggerFactory
 import java.util.concurrent.TimeUnit
 
 private val DEFAULT_INTERVAL = Duration.of(7, TimeUnit.DAYS)
 private const val DEFAULT_MAX_READ_INDICES = 2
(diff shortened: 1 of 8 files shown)


malish8632 (approved):
LGTM

heroic-elasticsearch-utils/src/main/java/com/spotify/heroic/elasticsearch/index/RotatingIndexMapping.kt
```diff
@@ -58,6 +61,27 @@ data class RotatingIndexMapping(
...
+        }
+
+        return indices
+    }
+
+    private fun dynamicMaxIndicesCount(now: Long, start: Long): Long {
+        val r = now - start
+
+        // Query indices within range + 1 to account for caching edge cases
+        return (r / interval) + 1
```
malish8632, line 82: What is our initial interval will be?

adsail: 2 days

PR #749, opened 2021-01-26
Remove legacy source input, optional metricType query input for testing

lmuhlha:
`source` was a legacy query option left over from when `events` were partially supported: 
https://github.com/spotify/heroic/commit/bc533078fad161efa76001cdc4f2a6e22dcb88d5#diff-0e5dda61cc1e44c094fed3ce698715de1d3b11747d9f312df0fbf6a52f70f1fcR222

This PR gets rid of source as an input and essentially re-names it to metricType with a few updates.

Since end users don't care about what `metricType` is being used, Distributions can instead set `metricType` based on what aggregation (tdigest) is being used.

### DIFF
--- a/heroic-core/src/main/java/com/spotify/heroic/CoreQueryManager.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/CoreQueryManager.java
@@ -35,7 +35,9 @@ import com.spotify.heroic.aggregation.AggregationInstance;
 import com.spotify.heroic.aggregation.BucketStrategy;
 import com.spotify.heroic.aggregation.DistributedAggregationCombiner;
 import com.spotify.heroic.aggregation.Empty;
+import com.spotify.heroic.aggregation.GroupInstance;
 import com.spotify.heroic.aggregation.TDigestAggregationCombiner;
+import com.spotify.heroic.aggregation.simple.TdigestInstance;
 import com.spotify.heroic.cache.QueryCache;
 import com.spotify.heroic.cluster.ClusterManager;
 import com.spotify.heroic.cluster.ClusterNode;
@@ -100,6 +102,7 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class CoreQueryManager implements QueryManager {
+
     private static final Logger log = LoggerFactory.getLogger(CoreQueryManager.class);
     public static final long SHIFT_TOLERANCE = TimeUnit.MILLISECONDS.convert(10, TimeUnit.SECONDS);
     public static final QueryTrace.Identifier QUERY_SHARD =
@@ -171,7 +174,7 @@ public class CoreQueryManager implements QueryManager {
         return expressions.get(0).eval(scope).visit(new Expression.Visitor<QueryBuilder>() {
             @Override
             public QueryBuilder visitQuery(final QueryExpression e) {
-                final Optional<MetricType> source = e.getSource();
+                final Optional<MetricType> metricType = e.getMetricType();
 
                 final Optional<QueryDateRange> range =
                     e.getRange().map(expr -> expr.visit(new Expression.Visitor<QueryDateRange>() {
@@ -202,9 +205,9 @@ public class CoreQueryManager implements QueryManager {
                 final Optional<Filter> filter = e.getFilter();
 
                 return new QueryBuilder()
-                    .source(source)
                     .range(range)
                     .aggregation(aggregation)
+                    .metricType(metricType)
                     .filter(filter);
             }
         });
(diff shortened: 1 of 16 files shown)


ao2017:
@lmuhlha   
There is a concept of metric_type downstream that is completely different from the metric_type as used in this PR.
metric_type upstream  is ( histogram, counter, gauge ....)
metric_type in heroic  is ( Point , Group, DistributionPoint ...)

Unfortunately, in heroic api this 2 concepts co-exist. In the query you can make the distinction by using "source". If you replace source  with metric_type, how are we going to ensure this change doesn't break anything?

I wanted to check but  this branch is not compiling. Check this  class  => **/Users/adeleo/src/heroic/heroic/heroic-dist/src/test/java/com/spotify/heroic/LoggingMetricModule.kt:** (36, 19) .

lmuhlha:
@ao2017 I fixed a few issues, now I'm just having some trouble with the tests. When i step through the code, the check for the tDigest aggregation instance sets the metricType properly. I noticed that there are tests that tDigest still works with no aggregation set. Is that how this is supposed to work? I think I'm a bit confused on functionality.

lmuhlha:
@ao2017  as far as the naming, I just figured metricType matched what it is in Heroic. I don't think it will conflict with anything in semantic metrics. We can come up with another name if need be, i just think `source` was confusing as well.

ao2017:
> @ao2017 as far as the naming, I just figured metricType matched what it is in Heroic. I don't think it will conflict with anything in semantic metrics. We can come up with another name if need be, i just think `source` was confusing as well.

I am not sure why the name source was used :). If everything is working as expected I am good.

ao2017:
> @ao2017 I fixed a few issues, now I'm just having some trouble with the tests. When i step through the code, the check for the tDigest aggregation instance sets the metricType properly. I noticed that there are tests that tDigest still works with no aggregation set. Is that how this is supposed to work? I think I'm a bit confused on functionality.

Let's look into together. I will ping you.

ao2017:
> > @ao2017 I fixed a few issues, now I'm just having some trouble with the tests. When i step through the code, the check for the tDigest aggregation instance sets the metricType properly. I noticed that there are tests that tDigest still works with no aggregation set. Is that how this is supposed to work? I think I'm a bit confused on functionality.
> 
> Let's look into together. I will ping you.

I checked. The refactoring on the consumer broke the ability to read distributionPoint from bigtable. It is causing query and consumer IT tests to fail. So the issue is not related to aggregation.

heroic-core/src/main/java/com/spotify/heroic/CoreQueryManager.java
```diff
@@ -278,20 +282,26 @@ public Group(List<ClusterShard> shards) {
...
                     () -> BucketStrategy.START));
 
-            final AggregationCombiner combiner;
-
+            if (aggregationInstance instanceof GroupInstance) {
+                if (((GroupInstance) aggregationInstance).getEach() instanceof TdigestInstance) {
+                    metricType = MetricType.DISTRIBUTION_POINTS;
+                }
+            }
+            if (aggregationInstance instanceof TdigestInstance) {
```
ao2017, line 290: This is not necessary. TdigestPoint is used in the final stage of aggregation to compute stat.

heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java
```diff
@@ -322,10 +323,11 @@ public QueryTrace buildTrace() {
 
                         fetchSeries.addAnnotation(series.toString());
                         fetches.add(() -> metricBackend.fetch(
-                            new FetchData.Request(source, series, range, options),
+                            new FetchData.Request(series, range, options),
```
ao2017, line 326: Why remove metric_type  in the request ? I think this is the root cause of the IT tests failure. 
Make sure every fetchData request is not defaulting to metric_type=Point

heroic-dist/src/test/java/com/spotify/heroic/AbstractConsumerIT.java
```diff
@@ -201,15 +203,16 @@ public void consumeManyMessages() throws Exception {
...
                         QueryOptions.defaults());
                 return coreComponent
                     .metricManager()
                     .useDefaultGroup()
                     .fetch(fetchDataRequest,
                         FetchQuotaWatcher.NO_QUOTA,
                         data::add,
-                        BlankSpan.INSTANCE
+                        BlankSpan.INSTANCE,
+                        MetricType.DISTRIBUTION_POINTS
```
ao2017, line 215: This test will fail unless you use metricType.POINT.

heroic-dist/src/test/java/com/spotify/heroic/AbstractConsumerIT.java
```diff
@@ -104,15 +104,16 @@ public void consumeOneMessage() throws Exception {
...
```
ao2017, line 116: Use MetricType.POINT instead

ao2017 (approved):

PR #751, opened 2021-02-02
improved Mandatory ID Filter to return info upon rejection

sming:
* it now correctly displays a JSON error message when the request is anonymous.
* also refactored MandatoryClientId & Shutdown Filters using the Template Method design pattern as they were
very similar.

### Note for reviewers

Checkout out [the Template Method design pattern](https://refactoring.guru/design-patterns/template-method) before reviewing this PR 👍🏻

### DIFF
--- a/heroic-core/src/main/java/com/spotify/heroic/http/HttpServer.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/http/HttpServer.java
@@ -213,7 +213,7 @@ public class HttpServer implements LifeCycles {
 
         context.addServlet(jerseyServlet, "/*");
         context.addFilter(new FilterHolder(new ShutdownFilter(stopping, mapper)), "/*", null);
-        context.addFilter(new FilterHolder(new MandatoryClientIdFilter(mapper)), "/*", null);
+        context.addFilter(new FilterHolder(new MandatoryClientIdFilter(mapper)), "/query/*", null);
         context.setErrorHandler(new JettyJSONErrorHandler(mapper));
 
         final RequestLogHandler requestLogHandler = new RequestLogHandler();
--- a/heroic-core/src/main/java/com/spotify/heroic/http/HttpServer.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/http/HttpServer.java
@@ -128,7 +128,7 @@ public class HttpServer implements LifeCycles {
             connectors.stream().map(c -> c.setup(server, address)).iterator());
     }
 
-    private ServerConnector findServerConnector(Server server) {
+    private static ServerConnector findServerConnector(Server server) {
         final Connector[] connectors = server.getConnectors();
 
         if (connectors.length == 0) {
@@ -213,7 +213,7 @@ public class HttpServer implements LifeCycles {
 
         context.addServlet(jerseyServlet, "/*");
         context.addFilter(new FilterHolder(new ShutdownFilter(stopping, mapper)), "/*", null);
-        context.addFilter(new FilterHolder(new MandatoryClientIdFilter()), "/*", null);
+        context.addFilter(new FilterHolder(new MandatoryClientIdFilter(mapper)), "/*", null);
         context.setErrorHandler(new JettyJSONErrorHandler(mapper));
 
         final RequestLogHandler requestLogHandler = new RequestLogHandler();
@@ -230,7 +230,7 @@ public class HttpServer implements LifeCycles {
         return handlers;
     }
 
-    private void makeRewriteRules(RewriteHandler rewrite) {
+    private static void makeRewriteRules(RewriteHandler rewrite) {
         {
             final RewritePatternRule rule = new RewritePatternRule();
             rule.setPattern("/metrics");
--- a/heroic-core/src/main/java/com/spotify/heroic/servlet/MandatoryClientIdFilter.java
+++ b/heroic-core/src/main/java/com/spotify/heroic/servlet/MandatoryClientIdFilter.java
@@ -21,32 +21,36 @@
 
 package com.spotify.heroic.servlet;
 
+import com.fasterxml.jackson.databind.ObjectMapper;
 import com.google.common.base.Strings;
+import com.spotify.heroic.ws.ErrorMessage;
+import com.spotify.heroic.ws.MandatoryClientIdErrorMessage;
 import java.io.IOException;
-import javax.servlet.Filter;
 import javax.servlet.FilterChain;
-import javax.servlet.FilterConfig;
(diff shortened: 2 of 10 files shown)


lmuhlha (dismissed):
worked like a charm locally:
```{"message":"Anonymous requests are not permitted","reason":"Service Unavailable","status":503,"type":"authentication-error"}%```
nice :)

File: heroic-core/src/main/java/com/spotify/heroic/http/HttpServer.java [hunk 1]
```diff
@@ -213,7 +213,7 @@ private HandlerCollection setupHandler() throws Exception {
 
         context.addServlet(jerseyServlet, "/*");
         context.addFilter(new FilterHolder(new ShutdownFilter(stopping, mapper)), "/*", null);
-        context.addFilter(new FilterHolder(new MandatoryClientIdFilter()), "/*", null);
+        context.addFilter(new FilterHolder(new MandatoryClientIdFilter(mapper)), "/*", null);
```
samfadrigalan, line 216: Do we want to also require client ids for non `/query/*` endpoints?

heroic-core/src/main/java/com/spotify/heroic/http/HttpServer.java
```diff
@@ -230,7 +230,7 @@ private HandlerCollection setupHandler() throws Exception {
         return handlers;
     }
 
-    private void makeRewriteRules(RewriteHandler rewrite) {
+    private static void makeRewriteRules(RewriteHandler rewrite) {
```
samfadrigalan, line 233: Why is this static?

heroic-core/src/main/java/com/spotify/heroic/http/HttpServer.java
lmuhlha: I believe we only want it for users querying metrics. Writes don't need a client id since the tags should be enough to tell you where it came from if that's the concern.

lmuhlha:
One question I had, why the switch from 400 to 503?

heroic-core/src/main/java/com/spotify/heroic/servlet/MandatoryClientIdFilter.java
```diff
@@ -21,32 +21,36 @@
...
-// method, else the JavaDoc for it doesn't render. Weird.
 @SuppressWarnings("checkstyle:LineLength")
-public class MandatoryClientIdFilter implements Filter {
+public class MandatoryClientIdFilter extends SimpleFilter {
 
     public static final String X_CLIENT_ID_HEADER_NAME = "X-Client-Id";
+    public static final String ERROR_MESSAGE_TEXT =
+            "This anonymous request has been rejected. Please add a 'x-client-id' " +
+                    "HTTP header to your request. Help can be found in the" +
+                    " #monitoring-users Slack channel.";
```
samfadrigalan, line 49: This is on the public repo. We probably don't want Spotify specific messages.

heroic-core/src/main/java/com/spotify/heroic/servlet/MandatoryClientIdFilter.java
```diff
@@ -62,27 +66,28 @@ public void init(FilterConfig filterConfig) throws ServletException { /* intenti
...
-        if (passesFilter(request)) {
-            chain.doFilter(request, response);
-        } else {
-            var httpResponse = HttpServletResponse.class.cast(response);
-            httpResponse.setStatus(Status.BAD_REQUEST.getStatusCode());
-        }
+
+        final var info =
+                new MandatoryClientIdErrorMessage("Anonymous requests are not permitted",
+                        Status.SERVICE_UNAVAILABLE);
```
samfadrigalan, line 75: Is this supposed to be `Status.BAD_REQUEST`?

heroic-core/src/main/java/com/spotify/heroic/ws/MandatoryClientIdErrorMessage.java
```diff
@@ -0,0 +1,34 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package com.spotify.heroic.ws;
+
+import javax.ws.rs.core.Response;
+
+public class MandatoryClientIdErrorMessage extends ErrorMessage {
+    public MandatoryClientIdErrorMessage(final String message, final Response.Status status) {
```
samfadrigalan, line 27: Would there be any case where the Response.Status where client id errors wouldn't be 400? If not, maybe we can just hard code 400 here  like
```
public MandatoryClientIdErrorMessage(final String message) {
    super(message, Status.BAD_REQUEST);
}
```

heroic-core/src/main/java/com/spotify/heroic/http/HttpServer.java
sming: I think that makes sense. I also have the sneaky feeling that writes haven't been tested at the system level and we really don't want those to fail...

heroic-core/src/main/java/com/spotify/heroic/http/HttpServer.java
sming: because static methods are easier to test, easier to reason about and faster to execute.

sming:
> One question I had, why the switch from 400 to 503?
that is a very good question, it should be 400. 503 is wrong.

heroic-core/src/main/java/com/spotify/heroic/servlet/MandatoryClientIdFilter.java
sming: good point.

heroic-core/src/main/java/com/spotify/heroic/servlet/MandatoryClientIdFilter.java
sming: Yes! That explains the 503 prolly.

heroic-core/src/main/java/com/spotify/heroic/ws/MandatoryClientIdErrorMessage.java
sming: good call

sming:
Hey @lmuhlha , @samfadrigalan could one of you please approve - I addressed all the issues raised. Cheers.

adsail:
@sming One quick thought -- It might be a good idea to update the public docs for [/query/metrics](https://spotify.github.io/heroic/docs/api/post-query-metrics) to stipulate the header requirement. It's currently part of the sample cURL but not called out as a requirement.

samfadrigalan (approved):

PR #756, opened 2021-02-08
Move from circle to github actions

lmuhlha:


sming:
Thanks for jumping on this @lmuhlha 🤜🏻 !

PR #763, opened 2021-03-01
Implement metric and log lines when a Bigtable request times out, per x-client-id

sming:
#### We need to know how many requests are hitting Bigtable timeouts per x-client-id

- I am a prism engineer
- Who wants to know the impact of his/her changes on our user base
- So that I can put out any fires efficiently, inform the user base if necessary, and protect heroic

#### Proposed Solution

- decorate the exception-handling at the API request level: add a check for the Bigtable timeout exception string
- if found, increment the according counter for the request's x-client-id
- if found, emit a log message that details the x-client-id (and the other usual exception info)

#### Design & Implementation Notes

nothing unusual. Just added : 
1. [detection logic](https://github.com/spotify/heroic/blob/8f1b4492f56110693e99ddc39d0cb64dde3ed7d9/heroic-core/src/main/java/com/spotify/heroic/metric/LocalMetricManager.java#L96), 
2. [a new metric e.g. `query-metrics-client-id-count-playerapp-timeout`](https://github.com/spotify/heroic/blob/33bf38ffa1a7778cfb2af279348c5253a78a671b/statistics/semantic/src/main/java/com/spotify/heroic/statistics/semantic/SemanticMetricBackendReporter.java#L155)
3. [logging e.g. `"bigtable-query-timeout...`](https://github.com/spotify/heroic/blob/8f1b4492f56110693e99ddc39d0cb64dde3ed7d9/heroic-core/src/main/java/com/spotify/heroic/querylogging/Slf4jQueryLogger.java#L89)

### DIFF
--- a/heroic-test/src/main/java/com/spotify/heroic/test/AbstractMetadataBackendIT.java
+++ b/heroic-test/src/main/java/com/spotify/heroic/test/AbstractMetadataBackendIT.java
@@ -378,7 +378,7 @@ public abstract class AbstractMetadataBackendIT {
     private void retrySome(final ThrowingRunnable action) throws Exception {
         AssertionError error = null;
 
-        for (int i = 0; i < 10; i++) {
+        for (int i = 0; i < 50; i++) {
             try {
                 action.run();
             } catch (final AssertionError e) {
--- a/heroic-component/src/main/java/com/spotify/heroic/common/FailureType.java
+++ b/heroic-component/src/main/java/com/spotify/heroic/common/FailureType.java
@@ -1,12 +1,12 @@
 /*
- * Copyright (c) 2019 Spotify AB.
+ * Copyright (c) 2015 Spotify AB.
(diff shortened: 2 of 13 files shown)


heroic-test/src/main/java/com/spotify/heroic/test/AbstractMetadataBackendIT.java
```diff
@@ -378,7 +378,7 @@ public void filterTest() throws Exception {
     private void retrySome(final ThrowingRunnable action) throws Exception {
         AssertionError error = null;
 
-        for (int i = 0; i < 10; i++) {
+        for (int i = 0; i < 50; i++) {
```
sming, line 381: Note for reviewers: this was timing out in CI so I increased. This won't be any slower, it's just more permissive.

malish8632 (approved):
LGTM

Issue #754, opened 2021-02-04
add x-client-id to markdown documentation examples for /query/[metrics|batch]


sming:
basically need to find a neat way of adding "x-client-id: my-app" to the example cURLs given in https://spotify.github.io/heroic/docs/api/post-query-metrics and https://spotify.github.io/heroic/docs/api/post-query-batch .

it's not that easy since we want this only to be added to these two endpoints. Probably need to create a "sub class" of whatever spits out the `curl -XPOST ...` text.

PR #766, opened 2021-03-24
adds x-client-id header to example curl in docco

sming:
addresses [add x-client-id to markdown documentation examples for /query/...](https://github.com/spotify/heroic/issues/754)

### DIFF
--- a/docs/_layouts/api-endpoint.html
+++ b/docs/_layouts/api-endpoint.html
@@ -48,7 +48,7 @@ layout: sidebar
         {%
         endunless %}
           {%- endcapture %}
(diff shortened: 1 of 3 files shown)


docs/_layouts/api-endpoint.html
```diff
@@ -48,7 +48,7 @@ <h4>Example Curl</h4>
         {%
         endunless %}
           {%- endcapture %}
-          $ curl -X{{ page.method }} {{ headers -}}
+          $ curl {{ page.x-client-id-header }} -X{{ page.method }} {{ headers -}}
```
lmuhlha, line 51: does `{{ headers -}}` not imply all headers here anyway?

